Issue #2, opened 2022-11-23
Bundler mnemonic and private key


Vid201:
### Problem

Currently, the bundler binary accepts a mnemonic folder or mnemonic file for the bundler's account. This can be confusing and should always be the same.

### Proposed solution

1. Implement standalone binary for creating new mnemonic and saving it to file.
2. Change the bundler to always accept the mnemonic file.

PR #16, opened 2022-11-24
Bundler wallet (#2)

Vid201:


bin/create-wallet.rs
```diff
@@ -0,0 +1,25 @@
...
+use clap::Parser;
+use expanded_pathbuf::ExpandedPathBuf;
+
+#[derive(Parser)]
+#[clap(
+    name = "aa-bundler-create-wallet",
+    about = "Bundler's wallet creation for EIP-4337 Account Abstraction"
+)]
+pub struct Opt {
+    #[clap(long, default_value = "./src/res/bundler")]
```
zsluedem, line 12: I think it would be better to use `$HOME/.aa-bundler` or just current dir for default_value.

bin/bundler.rs
```diff
@@ -52,19 +49,10 @@ fn main() -> Result<()> {
...
-                        .phrase(mnemonic_file.to_path_buf())
-                        .build()?
-                } else {
-                    let mut rng = rand::thread_rng();
-                    MnemonicBuilder::<English>::default()
-                        .write_to(opt.mnemonic_folder.to_path_buf())
-                        .build_random(&mut rng)?
-                };
+                let wallet = Wallet::from_file(opt.mnemonic_file);
+                println!("{:?}", wallet.signer);
```
zsluedem, line 53: I think it would be better to use `info!` to be consistent.

zsluedem (approved):
LGTM! Just left some nit comments.

zsluedem (approved):
Great~! :+1:

Issue #4, opened 2022-11-23
Connection to the EntryPoint smart contract


Vid201:
### Problem

Bundler must be able to conect to the EntryPoint smart contract and make calls.

### Proposed solution

Use `ethers-rs`.

zsluedem:
I want to take this one now. This relates to #1 .

zsluedem:
https://github.com/Vid201/aa-bundler/pull/25

PR #17, opened 2022-11-25
Generate type safe bindings for abstract accounts (#4)

zsluedem:
The first step for #4 .

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "thirdparty/account-abstraction"]
+	path = thirdparty/account-abstraction
+	url = https://github.com/eth-infinitism/account-abstraction.git
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -23,7 +23,7 @@ dependencies = [
  "dirs",
  "educe",
  "ethereum-interfaces",
- "ethereum-types 0.14.0",
+ "ethereum-types",
  "ethers",
  "expanded-pathbuf",
  "jsonrpsee",
@@ -52,9 +52,9 @@ dependencies = [
 
 [[package]]
 name = "aho-corasick"
-version = "0.7.19"
+version = "0.7.20"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b4f55bd91a0978cbfd91c457a164bab8b4001c833b7f323132c0a4e1922dd44e"
+checksum = "cc936419f96fa211c1b9166887b38e5e40b19958e5b895be7c1f93adec7071ac"
 dependencies = [
  "memchr",
 ]
@@ -76,6 +76,15 @@ name = "arrayvec"
 version = "0.7.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "8da52d66c7071e2e3fa2a1e5c6d088fec47b593032b254f5e980de8ea54454d6"
+
+[[package]]
+name = "ascii-canvas"
+version = "3.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8824ecca2e851cec16968d54a01dd372ef8f95b244fb84b84e70128be347c3c6"
+dependencies = [
+ "term",
+]
 
 [[package]]
 name = "async-stream"
@@ -126,7 +135,7 @@ version = "0.2.14"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d9b39be18770d11421cdb1b9947a45dd3f37e93092cbf377614828a319d5fee8"
 dependencies = [
- "hermit-abi",
+ "hermit-abi 0.1.19",
  "libc",
  "winapi",
 ]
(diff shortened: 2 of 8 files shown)


Cargo.lock
```diff
@@ -3514,48 +3838,24 @@ version = "0.42.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "41d2aa71f6f0cbe00ae5167d90ef3cfe66527d6f613ca78ac8024c3ccab9a19e"
 
-[[package]]
-name = "windows_aarch64_msvc"
-version = "0.36.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9bb8c3fd39ade2d67e9874ac4f3db21f0d710bee00fe7cab16949ec184eeaa47"
```
zsluedem, line 3521: @Vid201 Are you using window for developing. I am not sure if these changes would break anything for you. Please check.

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "thirdparty/account-abstraction"]
+	path = thirdparty/account-abstraction
+	url = https://github.com/eth-infinitism/account-abstraction.git
```
zsluedem, line 3: @Vid201 I am using submodules to keep track of the interfaces here. I am not sure whether you prefer submodule or subtree. I am fine with either way. Please tell me what you prefer.

src/contracts/mod.rs
```diff
@@ -0,0 +1,29 @@
...
+abigen!(
+    StakeManager,
+    "$OUT_DIR/IStakeManager.sol/IStakeManager.json"
+);
+
+abigen!(Wallet, "$OUT_DIR/IWallet.sol/IWallet.json");
+
+abigen!(
+    UserOperation,
+    "$OUT_DIR/UserOperation.sol/UserOperationLib.json"
```
zsluedem, line 28: @Vid201 Just FYI. Here it would generate the same data structure `UserOperation`.

Cargo.lock
Vid201: I am using Linux. I think we should add Cargo.lock to .gitignore in fact.

.gitmodules
Vid201: submodules are great!

src/contracts/mod.rs
Vid201: Hmm, ok. Then we can import this data structure in `user_operation.rs` (and implement additional things that will be needed).

Cargo.toml
```diff
@@ -33,6 +33,8 @@ tracing-subscriber = "0.3"
 protobuf-src = "1.1.0"
 prost-build = "0.11"
 tonic-build = "0.8"
+anyhow = "1"
```
Vid201, line 36: anyhow and ethers are now twice included as deps

build.rs
```diff
@@ -14,6 +19,27 @@ fn make_protos(protos: &[&str]) {
...
+        .artifacts(target_path)
+        .build_infos(root.join("contracts").join("build-info"))
+        .root(root)
+        .build()?;
+    let project = Project::builder().paths(build_path_config).build()?;
+    project.rerun_if_sources_changed();
+    let compiled = project.compile()?;
+    assert!(
+        !compiled.has_compiler_errors(),
+        "Compiling abstract account interfaces failed: {:?}",
```
Vid201, line 37: ... EIP-4337 interfaces ...

build.rs
```diff
@@ -24,4 +50,6 @@ fn main() {
     ];
 
     make_protos(&protos);
+
+    compile_aa_interfaces().expect("Compiling abstract account interfaces should pass.");
```
Vid201, line 54: ... EIP-4337 interfaces ...

build.rs
```diff
@@ -14,6 +19,27 @@ fn make_protos(protos: &[&str]) {
         .unwrap();
 }
 
+fn compile_aa_interfaces() -> anyhow::Result<()> {
+    let root = PathBuf::from(env!("CARGO_MANIFEST_DIR")).join("thirdparty/account-abstraction");
+    let target_path = PathBuf::from(env::var("OUT_DIR").unwrap());
+    let build_path_config = ProjectPathsConfig::builder()
+        // current only interfaces are what we need
```
Vid201, line 26: "only interfaces are needed"

sorry for complicating so much about comments :sweat_smile:

zsluedem: It is OK~! I am not a native English speaker. Don't hesitate to comment on those words as long as it could help people to understand. I can benefit myself from your comments, too.

Cargo.lock
zsluedem: I am not sure, actually. I think we could keep it.
I just googled with some [reference](https://doc.rust-lang.org/cargo/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries).

Cargo.toml
zsluedem: This is for the `build.rs` only. Build.rs could only be imported in `[build-dependencies]`

zsluedem: The `anyhow` in `[dependencies]` doesn't help in `build.rs`

Vid201: Yeah, my bad, looked at it too fast.

Vid201: Then we can alphabetically sort them so it's easier to find dep later.

Cargo.lock
Vid201: Ok, you're right. Though I saw some projects removing it.

Nevertheless, it doesn't break anything for me, so it's fine.

build.rs
```diff
@@ -1,4 +1,9 @@
-use std::{env, path::PathBuf};
+use std::{
+    env,
+    path::{Path, PathBuf},
```
Vid201, line 3: Path - unused import

Vid201:
@zsluedem I think we should add info about installing solc to README.md (with link to https://docs.soliditylang.org/en/v0.8.9/installing-solidity.html and version required).

README.md
```diff
@@ -10,6 +10,10 @@
 
 For more information: https://hackmd.io/@Vid201/aa-bundler-rust
 
+## Prerequisites
+
+1. [install solc](https://docs.soliditylang.org/en/v0.8.17/installing-solidity.html) >=0.8.12
```
zsluedem, line 15: @Vid201 I am curious whether `protoc` is needed. I tried and it seems unnecessary.

Vid201: It's not necessary, no need to add that.

Vid201:
LGTM! :rocket:

Vid201 (approved):

Issue #7, opened 2022-11-23
In-memory user operation pool


Vid201:
Received user operations are persisted in memory untill they are included in the bundle.

PR #19, opened 2022-11-28
User operation pack and hash (#7)

Vid201:


Vid201:
This PR can close #22 , I would like to push that to main before continuing with #7 (I will continue then on the same PR). @zsluedem

src/types/user_operation.rs
```diff
@@ -17,3 +22,161 @@ pub struct UserOperation {
...
...ons[1].pack(), String::from("0x000000000000000000000000663f3ad617193148711d28f5334ee4ed070166020000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000016000000000000000000000000000000000000000000000000000000000000001800000000000000000000000000000000000000000000000000000000000030d4000000000000000000000000000000000000000000000000000000000000186a0000000000000000000000000000000000000000000000000000000000000520800000000000000000000000000000000000000000000000000000000b2d05e00000000000000000000000000000000000000000000000000000000003b9aca0000000000000000000000000000000000000000000000000000000000000001a000000000000000000000000000000000000000000000000000000000000001c000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000417cb39607585dee8e297d0d7a669ad8c5e43975220b6773c10a138deadbc8ec864981de4b9b3c735288a217115fb33f8326a61ddabc60a534e3b5536515c70f931c00000000000000000000000000000000000000000000000000000000000000").into_bytes());
```
zsluedem, line 94: I am just curious here. Where are these test cases (and below test cases) coming from? I was looking at the typescript aa repo but didn't find.

src/types/user_operation.rs
```diff
@@ -17,3 +22,161 @@ pub struct UserOperation {
     pub paymaster_and_data: Bytes,
     pub signature: Bytes,
 }
+
+impl UserOperation {
+    pub fn pack(&self) -> Vec<u8> {
+        self.clone().encode_hex().into_bytes()
+    }
+
+    pub fn pack_for_signature(&self) -> Vec<u8> {
```
zsluedem, line 31: This is the rust implementation of https://github.com/eth-infinitism/account-abstraction/blob/develop/contracts/interfaces/UserOperation.sol#L57-L71, right?

zsluedem (commented):
For [d610bfb](https://github.com/Vid201/aa-bundler/pull/19/commits/d610bfbd1eada9f357280143b9307dd4364ffa93) .
This is personal preference. I think it would be better to use `rebase` instead of `merge` the branch which could end up with a less merge knot in the git history. What do you think?

src/types/user_operation.rs
Vid201: Yes. You can find more here: https://github.com/eth-infinitism/account-abstraction/blob/develop/test/UserOp.ts#L35

src/types/user_operation.rs
Vid201: First test case: https://github.com/eth-infinitism/account-abstraction/blob/develop/test/UserOp.ts#L113
Second test case is taken from this script: https://github.com/eth-infinitism/account-abstraction/blob/develop/test/simple-wallet.test.ts#L67

Vid201:
> For [d610bfb](https://github.com/Vid201/aa-bundler/pull/19/commits/d610bfbd1eada9f357280143b9307dd4364ffa93) . This is personal preference. I think it would be better to use `rebase` instead of `merge` the branch which could end up with a less merge knot in the git history. What do you think?

Yeah, that makes sense. Good suggestion!

src/types/user_operation.rs
zsluedem: Thanks~! I figure that out now~! Good work on this

zsluedem (approved):
Good work.

PR #21, opened 2022-11-30
Add bundler config

zsluedem:


src/bundler/bundler.rs
```diff
@@ -1,5 +1,37 @@
+use std::str::FromStr;
+
+use clap::Parser;
+use ethereum_types::U256;
+use ethers::types::Address;
+
 use crate::models::wallet::Wallet;
 
+#[derive(Debug, Parser, PartialEq)]
+pub struct BundlerOpts {
```
Vid201, line 10: I think this should be only Opts, we using full import path in the bundler.rs: aa_bundler::bundler::bundler::Opts

src/bundler/bundler.rs
```diff
@@ -1,5 +1,37 @@
+use std::str::FromStr;
+
+use clap::Parser;
+use ethereum_types::U256;
```
Vid201, line 4: I think we should switch to ethers::type fully, I had some problems with ethereum_types in the past (ethers::types::U256)

src/bundler/bundler.rs
```diff
@@ -9,3 +41,39 @@ impl Bundler {
...
+
+#[cfg(test)]
+mod test {
+    use super::BundlerOpts;
+    use clap::Parser;
+    use ethereum_types::{Address, U256};
+    use std::str::FromStr;
+
+    #[test]
+    fn test_bundle_opt() {
```
Vid201, line 53: This is my personal preference, but I do not write prefix `test_`before test names. What do you think? We should have the same naming across codebase (decide for the one).

bin/bundler.rs
```diff
@@ -31,6 +31,13 @@ pub struct Opt {
 
     #[clap(long, default_value = "127.0.0.1:4337")]
     pub rpc_listen_address: String,
+
+    // execution client rpc endpoint
+    #[clap(long)]
+    pub network: SocketAddr,
```
Vid201, line 37: `pub eth_client_address: SocketAddr` would be better name IMO

src/bundler/bundler.rs
zsluedem: > We should have the same naming across codebase (decide for the one).

Agree.

So you prefer just `bundle_opt`?

bin/bundler.rs
zsluedem: I copied the name from [typescript bundler](https://github.com/eth-infinitism/bundler/blob/main/packages/bundler/src/runBundler.ts#L88) and I agree that your name is better.

src/bundler/bundler.rs
Vid201: Yes.

Vid201 (approved):

PR #24, opened 2022-12-05
Sanity checks

Vid201:
This PR adds sanity checks for user operations (canonical mempool).

- [x] sender or init code
- [x] factory
- [x] verification gas
- [x] paymaster and data
- [x] call gas limit
- [x] max fee per gas
- [x] single user operation in the mempool

Sanity check verification will be abstracted in the future PR.


src/uopool/services/simulation.rs
```diff
@@ -0,0 +1,131 @@
...
+        let simulation_response = self
+            .entry_point
+            .simulate_validation(entry_point::UserOperation::from(user_operation.clone()))
+            .await;
+
+        let result = format!("{:?}", simulation_response);
+
+        if let Err(ContractError::MiddlewareError(ProviderError::JsonRpcClientError(
+            json_rpc_client_error,
+        ))) = simulation_response
```
zsluedem, line 53: I am also working on the API and I found the error handling is very bad in ethers-rs(I asked the ethers-rs team and they agreed :) ). This error could be deeply embedded if people choose different embedded middleware. I chose a different way to handle errors here. 
I just put a comment here to mark. We could revisit the difference later.

Vid201: Yeah, I agree, it's very bad.

It's great that you chose a different way; we can compare and see what's better. Calls to the entry-point smart contract should probably be moved to a separate file.

Cargo.toml
```diff
@@ -13,7 +13,7 @@ clap = { version = "4", features = ["derive"] }
 dirs = "4.0"
 educe = { version = "0.4", features = ["Debug", "Default"] }
 ethereum-interfaces = { git = "https://github.com/ledgerwatch/interfaces" }
-ethers = "1.0.0"
+ethers = { git = "https://github.com/gakonst/ethers-rs", branch = "master", features = ["ethers-solc"]  }
```
zsluedem, line 16: branch = "master" is not stable because master is always changing. We should use some rev directly until they release a new version.

src/chain/gas.rs
```diff
@@ -0,0 +1,92 @@
...
+impl Default for Overhead {
+    fn default() -> Self {
+        Self {
+            fixed: U256::from(21000),
+            per_user_op: U256::from(18300),
+            per_user_op_word: U256::from(4),
+            zero_byte: U256::from(4),
+            non_zero_byte: U256::from(16),
+            bundle_size: U256::from(1),
+            sig_size: U256::from(65),
```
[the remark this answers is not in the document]
zsluedem: Is it from https://github.com/eth-infinitism/bundler/blob/main/packages/sdk/src/calcPreVerificationGas.ts#L44-L52 ? Do you know why? I am just curious.

src/uopool/services/sanity_check.rs
```diff
@@ -0,0 +1,559 @@
...
+        self.sender_or_init_code(user_operation).await?;
+
+        // If initCode is not empty, parse its first 20 bytes as a factory address. Record whether the factory is staked, in case the later simulation indicates that it needs to be. If the factory accesses global state, it must be staked - see reputation, throttling and banning section for details.
+        self.verify_factory(user_operation, entry_point).await?;
+
+        // The verificationGasLimit is sufficiently low (<= MAX_VERIFICATION_GAS) and the preVerificationGas is sufficiently high (enough to pay for the calldata gas cost of serializing the UserOperation plus PRE_VERIFICATION_OVERHEAD_GAS)
+        self.verification_gas(user_operation)?;
+
+        // The paymasterAndData is either empty, or start with the paymaster address, which is a contract that (i) currently has nonempty code on chain, (ii) has a sufficient deposit to pay for the UserOperation, and (iii) is not currently banned. During simulation, the paymaster's stake is also checked, depending on its storage usage - see reputation, throttling and banning section for details.
+        self.verify_paymaster(user_operation, entry_point).await?;
```
zsluedem, line 279: It seems like `verify_paymaster` doesn't check the `(i)` currently.

src/uopool/services/sanity_check.rs
```diff
@@ -0,0 +1,559 @@
...
+        }
+
+        Ok(())
+    }
+
+    fn call_gas_limit(
+        &self,
+        user_operation: &UserOperation,
+    ) -> Result<(), BadUserOperationError<M>> {
+        let non_zero_value_call = gas::non_zero_value_call();
```
zsluedem, line 154: I thought it would require `eth_estimate` API.

src/chain/gas.rs
```diff
@@ -0,0 +1,92 @@
...
+use ethers::types::{u256_from_f64_saturating, U256};
+use std::ops::Deref;
+
+pub const WARM_STORAGE_READ_COST: u64 = 100; // EIP 2929
+pub const CALL_VALUE_TRANSFER_GAS: u64 = 9000; // Ethereum yellow paper
+pub const CALL_GAS: u64 = 700; // EIP 150
+pub const CALL_STIPEND: u64 = 2300; // Ethereum yellow paper
+
+pub fn non_zero_value_call() -> U256 {
+    U256::from(CALL_VALUE_TRANSFER_GAS + WARM_STORAGE_READ_COST + CALL_GAS + CALL_STIPEND)
```
zsluedem, line 11: Where could I find info about this calculation?

src/uopool/services/sanity_check.rs
```diff
@@ -0,0 +1,559 @@
...
+
+        if user_operation.max_priority_fee_per_gas > user_operation.max_fee_per_gas {
+            return Err(BadUserOperationError::HighMaxPriorityFeePerGas {
+                max_priority_fee_per_gas: user_operation.max_priority_fee_per_gas,
+                max_fee_per_gas: user_operation.max_fee_per_gas,
+            });
+        }
+
+        if base_fee_estimation + user_operation.max_priority_fee_per_gas
+            > user_operation.max_fee_per_gas
```
zsluedem, line 183: I don't understand the calculation here. Could you elaborate a little bit?

I thought it would be just
```rust
max_fee_per_gas > config.minimum_gas && max_fee_per_gas > base_fee
max_priority_fee_per_gas > config.minimum_gas && max_priority_fee_per_gas > base_fee
```
I don't quite understand the logic here.

src/uopool/services/uopool.rs
```diff
@@ -12,68 +17,111 @@ use crate::{
...
 use tonic::Response;
 
 pub type UoPoolError = ErrorObject<'static>;
 
-pub struct UoPoolService {
+pub struct UoPoolService<M: Middleware> {
+    pub entry_points: Arc<HashMap<MempoolId, EntryPoint<M>>>,
     pub mempools: Arc<RwLock<HashMap<MempoolId, MempoolBox<Vec<UserOperation>>>>>,
     pub reputations: Arc<RwLock<HashMap<MempoolId, ReputationBox<Vec<ReputationEntry>>>>>,
+    pub sanity_check_results: Arc<RwLock<HashMap<UserOperationHash, HashSet<SanityCheckResult>>>>,
```
zsluedem, line 39: Just curious. Why is the `sanity_check_results` needed?

zsluedem (commented):
Very cool implementation on https://github.com/eth-infinitism/account-abstraction/blob/develop/eip/EIPS/eip-4337.md#client-behavior-upon-receiving-a-useroperation. I left some comments.

src/chain/gas.rs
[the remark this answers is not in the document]
Vid201: It's from there, yes. Some values (like zeroByte and nonZeroByte) are just hardcoded gas values for evm opcodes (https://github.com/wolflo/evm-opcodes/blob/main/gas.md#appendix---dynamic-gas-costs), other values are based on the structure of user operations (some fields with the fixed size of bytes are always present).

src/uopool/services/sanity_check.rs
Vid201: Ah, missed that. Tnx

src/uopool/services/sanity_check.rs
Vid201: You are right, I will check this again. Maybe non zero value call is not even necessary.

src/chain/gas.rs
Vid201: Non zero value call: https://github.com/wolflo/evm-opcodes/blob/main/gas.md#aa-call-operations

src/uopool/services/sanity_check.rs
Vid201: What do you mean here by config.minimum_gas here? Bundler will decide if user operations pay enough fees solely on the current base fee in the network (if it makes sense to include this user operation based on the existing network conditions).

The Max priority fee per gas is the tip the user wants to pay the bundler to include his user operation. The sum of the base fee and tip should be bigger than the max fee per gas (ma fee is the maximum user will pay for user operation).

Your verification would not work because max priority is a much smaller number than the base fee (it's just a tip, it could be only 1 wei in theory: https://docs.alchemy.com/docs/maxpriorityfeepergas-vs-maxfeepergas).

Here we only check the combination of values that don't make sense, and that is if the max priority is bigger than the max fee and if the sum of the base fee and max priority is bigger than the max fee (which means the user operation is offering too low fee to be included).

src/uopool/services/uopool.rs
Vid201: Sanity check results are needed for the simulation. During the simulation, we have to allow some OPCODES only if the paymaster is staked, which is verified here in sanity checks (so there is no need to check that again in simulation). Simulation is also performed two times, once when the user operation is received and again during the bundling.

zsluedem: Make perfect sense.

src/uopool/services/sanity_check.rs
zsluedem: `minimum_gas` means the minimum gas  the bundler want at least which based on the bundler. I am not sure whether I understand the spec correctly here but apparently, we have different understandings.

Let me revisit the spec again.

Vid201: Hmm, I checked the spec again. Maybe we are both right.

`The maxFeePerGas and maxPriorityFeePerGas are above a configurable minimum value that the client is willing to accept. At the minimum, they are sufficiently high to be included with the current block.basefee.`

The first sentence is what you meant, you're right, I missed that. But my understanding of the second sentence is the calculation in the code. So max fee is what use wants to pay maximum with base fee being burned, and max priority being a tip (so base fee + tip must be less or equal to max fee).

zsluedem: Cool. I get it now. We are both right:)

Vid201: For now, I added only configurable min maxPriorityFeePerGas since it seems more intuitive for me to have it configurable. I think we can add configurable min maxFeePerGas later.

src/uopool/database.rs
```diff
@@ -131,6 +131,21 @@ impl<E: EnvironmentKind> Mempool for UserOpDatabase<E> {
             .unwrap_or_else(|_| vec![])
     }
 
+    fn get_number_by_sender(&self, sender: &Address) -> usize {
```
Vid201, line 134: @zsluedem Please check if this fun is ok.

src/uopool/services/uopool.rs
zsluedem: After some thinkings, I think it would be better to pipeline the sanity check and validation together instead of storing the results here because it would keep consuming the memory(or you plan to remove it after the validation?). But I am cool with the current solution right now. Let's see what's better later.

zsluedem (approved):
Generally looks good. Thanks for addressing the comments. Good work.

src/uopool/services/uopool.rs
Vid201: Yeah, I was in a dilemma about what would be better.

It should be removed simultaneously as the user operation is removed from the mempool. Let's keep it as is, and we can address this later.

PR #25, opened 2022-12-19
Entry point api

zsluedem:
Introduce a wrapped entry point api that handle the error in our own way.

### DIFF
--- a/src/contracts/entrypoint.rs
+++ b/src/contracts/entrypoint.rs
@@ -61,19 +61,10 @@ impl<M: Middleware + 'static> EntryPoint<M> {
             })
     }
 
-    pub async fn get_user_op_hash<U: Into<UserOperation>>(
-        &self,
-        user_operation: U,
-    ) -> Result<[u8; 32], EntryPointErr> {
-        let res: Result<[u8; 32], ContractError<M>> =
-            self.api.get_user_op_hash(user_operation.into()).await;
-        res.map_err(|e| EntryPointErr::UnknownErr(format!("Get user op hash error with {:?}", e)))
-    }
-
     pub async fn simulate_validation<U: Into<UserOperation>>(
         &self,
         user_operation: U,
-    ) -> Result<SimulateResult, EntryPointErr> {
+    ) -> Result<SimulateValidationResult, EntryPointErr> {
         let request_result = self.api.simulate_validation(user_operation.into()).await;
         match request_result {
             Ok(_) => Err(EntryPointErr::UnknownErr(
@@ -86,10 +77,12 @@ impl<M: Middleware + 'static> EntryPoint<M> {
                         Err(EntryPointErr::FailedOp(failed_op))
                     }
                     entry_point_api::EntryPointAPIErrors::SimulationResult(res) => {
-                        Ok(SimulateResult::SimulationResult(res))
+                        Ok(SimulateValidationResult::SimulationResult(res))
                     }
                     entry_point_api::EntryPointAPIErrors::SimulationResultWithAggregation(res) => {
-                        Ok(SimulateResult::SimulationResultWithAggregation(res))
+                        Ok(SimulateValidationResult::SimulationResultWithAggregation(
+                            res,
+                        ))
                     }
                     _ => Err(EntryPointErr::UnknownErr(format!(
                         "Simulate validation with invalid error: {:?}",
@@ -166,7 +159,7 @@ pub enum EntryPointErr {
 }
 
 #[derive(Debug)]
-pub enum SimulateResult {
+pub enum SimulateValidationResult {
     SimulationResult(entry_point_api::SimulationResult),
     SimulationResultWithAggregation(entry_point_api::SimulationResultWithAggregation),
 }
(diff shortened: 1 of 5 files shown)


src/contracts/gen.rs
```diff
@@ -0,0 +1,47 @@
...
+);
+
+abigen!(Account, "$OUT_DIR/IAccount.sol/IAccount.json");
+
+abigen!(
+    UserOperation,
+    "$OUT_DIR/UserOperation.sol/UserOperationLib.json"
+);
+
+impl From<UserOp> for entry_point_api::UserOperation {
```
zsluedem, line 31: This is dup to the pr #24  . Just want to put it here to compile well. Should be removed later.

src/contracts/entrypoint.rs
```diff
@@ -0,0 +1,238 @@
...
+}
+
+#[derive(Debug)]
+pub enum SimulateResult {
+    SimulationResult(entry_point_api::SimulationResult),
+    SimulationResultWithAggregation(entry_point_api::SimulationResultWithAggregation),
+}
+
+#[derive(Debug, Deserialize, PartialEq)]
+pub struct JsonRpcError {
```
zsluedem, line 175: This is actually the same as https://github.com/gakonst/ethers-rs/blob/master/ethers-providers/src/transports/common.rs#L14-L22. But sadly, ethers-rs doesn't expose this type.

src/contracts/entrypoint.rs
```diff
@@ -0,0 +1,238 @@
...
+    pub message: String,
+    /// Additional data
+    pub data: Option<String>,
+}
+
+impl FromStr for JsonRpcError {
+    type Err = anyhow::Error;
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let re = Regex::new(
+            r###"code: (\d+), message: ([^,]*), data: (None|Some\(String\("([^)]*)"\))"###,
```
zsluedem, line 188: https://github.com/gakonst/ethers-rs/blob/master/ethers-providers/src/transports/common.rs#L26
Revert back to JsonRpcError based on the `Display`.

src/contracts/entrypoint.rs
```diff
@@ -0,0 +1,238 @@
...
+            }
+        }
+    }
+
+    async fn handle_aggregated_ops<U: Into<UserOperation>>(
+        &self,
+        ops_per_aggregator: Vec<U>,
+        beneficiary: Address,
+    ) -> Result<(), EntryPointErr> {
+        todo!()
```
zsluedem, line 156: I haven't found good test cases for this. I would make it work in another pr.

src/contracts/entrypoint.rs
```diff
@@ -0,0 +1,238 @@
...
+
+#[cfg(test)]
+mod test {
+    use ethers::types::Bytes;
+
+    use super::JsonRpcError;
+    use std::str::FromStr;
+
+    #[test]
+    fn json_rpc_errpr_parse() {
```
Vid201, line 213: `json_rpc_err_parse`?

src/contracts/gen.rs
```diff
@@ -0,0 +1,47 @@
+use crate::types::user_operation::UserOperation as UserOp;
+use ethers::contract::abigen;
+abigen!(
```
Vid201, line 3: Should we comment out contracts we don't need?

src/contracts/entrypoint.rs
Vid201: Do you know if there is any specific reason why that type is not exposed? Maybe we should update ethers-rs to expose it.

src/contracts/entrypoint.rs
```diff
@@ -0,0 +1,238 @@
...
+#[derive(Debug)]
+pub enum EntryPointErr {
+    FailedOp(entry_point_api::FailedOp),
+    NetworkErr, // TODO
+    DecodeErr(String),
+    UnknownErr(String), // describe impossible error. We should fix the codes here(or contract codes) if this occurs.
+}
+
+#[derive(Debug)]
+pub enum SimulateResult {
```
Vid201, line 169: SimulateValidationResult sounds better to me (more precise based on the SC's function name

src/contracts/entrypoint.rs
```diff
@@ -0,0 +1,238 @@
...
+                AbiDecode::decode_hex(data).map_err(|_| {
+                    EntryPointErr::DecodeErr(format!(
+                        "{:?} data field could not be deserialize to EntryPointAPIErrors",
+                        err_msg
+                    ))
+                })
+            })
+    }
+
+    pub async fn get_user_op_hash<U: Into<UserOperation>>(
```
Vid201, line 64: I do not think this function is needed; it's implemented in user_operation.rs (and no need to make another RPC call to provider/client).

Vid201 (changes requested):

Vid201:
This looks good; good job! Left some comments

src/contracts/entrypoint.rs
```diff
@@ -0,0 +1,238 @@
...
+        Ok(JsonRpcError {
+            code,
+            message: message.to_string(),
+            data,
+        })
+    }
+}
+
+#[cfg(test)]
+mod test {
```
Vid201, line 206: `mod tests`

src/contracts/entrypoint.rs
zsluedem: Besides JsonRpcError, there are a lot of middleware error types in ethers-rs. What kind of error you get depends on the middleware and it is dynamic dispatch. The ethers-rs team told me it is going to change in version 2. I didn't explore how difficult to fix it in the current version.

FYI, the current solution is not handling all the situations well. I haven't met any network errors (or middleware error) yet. And we need to improve this in the future.

Vid201: Ok, got it, thanks for the explanation

Vid201 (approved):

Vid201:
@zsluedem It's ready for merge unless you plan to make any additional changes.

zsluedem:
@Vid201 Thanks

PR #26, opened 2022-12-19
Validate tests

zsluedem:
** You have to install `Yarn` and `foundry` to make it work**

[Install foundry](https://github.com/foundry-rs/foundry)

This is the rust codes of validating tests in https://github.com/eth-infinitism/bundler/blob/main/packages/bundler/test/ValidateManager.test.ts .

1. The pr introduce foundry as a testing evm node which would provide enough function for bundler to interact with.
2. Include bundler repo
3. Introduces the validating test cases.
4. Enable clippy lint in tests.


tests/integration_test.rs
```diff
@@ -0,0 +1,304 @@
...
+    ) {
+        let user_op_hash = user_op.hash(entry_point_address, chain_id);
+        let mut eth_message = "\x19Ethereum Signed Message:\n32".to_string().into_bytes();
+        eth_message.extend_from_slice(user_op_hash.as_bytes());
+        let hash: H256 = keccak256(&eth_message).into();
+        let signature = key.sign_hash(hash);
+        user_op.signature = Bytes::from(signature.to_vec());
+    }
+
+    fn install_prerequisites() {
```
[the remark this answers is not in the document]
Vid201: Would be nice to change makefile: 

```
fetch-thirdparty:
	git submodule update --init && cd thirdparty-account-abstraction && yarn install && cd ../..
```

tests/integration_test.rs
```diff
@@ -0,0 +1,304 @@
...
+                .unwrap(),
+                Default::default(),
+            )
+            .await;
+        println!("{:?}", res);
+        Ok(())
+    }
+
+    #[tokio::test]
+    async fn interact_with_local() -> anyhow::Result<()> {
```
Vid201, line 193: This test does not work for me. I get `thread 'test::interact_with_local' panicked at 'couldnt start anvil: Os { code: 2, kind: NotFound, message: "No such file or directory" }', /home/vid/.cargo/registry/src/github.com-1ecc6299db9ec823/ethers-core-1.0.2/src/utils/anvil.rs:251:37`

Can the call to `setup()` (setting up anvil while it is already set) be the reason?

Vid201 (changes requested):

tests/integration_test.rs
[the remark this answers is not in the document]
zsluedem: good point.

tests/integration_test.rs
```diff
@@ -0,0 +1,304 @@
...
+            call_gas_limit: U256::from(200000),
+            verification_gas_limit: U256::from(100000),
+            pre_verification_gas: U256::from(21000),
+            max_fee_per_gas: U256::from(3000000000 as u64),
+            max_priority_fee_per_gas: U256::from(1000000000),
+            paymaster_and_data: Bytes::default(),
+            signature: Bytes::default(),
+        };
+
+        user_op.pack_for_signature();
```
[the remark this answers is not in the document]
zsluedem: yes

tests/integration_test.rs
```diff
@@ -0,0 +1,304 @@
...
+
+            let simple_acccount =
+                ContractFactory::new(simple_account.abi, simple_account.bytecode, client)
+                    .deploy(entry_point_contract.address())?
+                    .send()
+                    .await?;
+            Ok((entry_point_contract.address(), simple_acccount.address()))
+        }
+    }
+    pub fn sign(
```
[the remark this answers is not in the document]
zsluedem: You are right.

tests/integration_test.rs
Vid201: Yes, it's fine now. I tried again and it goes through now, weird. :slightly_smiling_face:

Vid201:
This is good, will be very helpful :rocket:

Cargo.toml
```diff
@@ -53,3 +54,7 @@ name = "bundler-rpc"
 [[bin]]
 path = "bin/create-wallet.rs"
 name = "create-wallet"
+
+[[bin]]
+path = "bin/compile_contracts.rs"
```
Vid201, line 59: Let's use '-' as with other bin files. `compile-contracts.rs`

tests/validate_tests.rs
```diff
@@ -0,0 +1,512 @@
...
+        verification_gas_limit: U256::from(1000000),
+        pre_verification_gas: U256::from(50000),
+        max_fee_per_gas: U256::from(0),
+        max_priority_fee_per_gas: U256::from(0),
+        paymaster_and_data: Bytes::from(paymaster_and_data),
+        signature,
+    }
+}
+
+fn validate(_user_op: UserOperation) -> anyhow::Result<()> {
```
zsluedem, line 210: @Vid201 You have to implement this method for validations you wrote in your [pr](https://github.com/Vid201/aa-bundler/pull/24).

tests/validate_tests.rs
```diff
@@ -0,0 +1,512 @@
...
+        init_code,
+        init_func,
+        context.opcodes_factory.address,
+    )
+    .await
+    .expect("succeed");
+}
+
+#[tokio::test]
+#[should_panic]
```
zsluedem, line 256: The `#[should_panic]` flag should be removed when the validating function is implemented. The case panic right now is because `validate` always return Ok

Makefile
```diff
@@ -15,13 +15,26 @@ format:
 
 lint:
 	cargo fmt --all -- --check
-	cargo clippy -- -D warnings -A clippy::derive_partial_eq_without_eq
+	cargo clippy --tests -- -D warnings -A clippy::derive_partial_eq_without_eq
 
 build:
+	make fetch-thirdparty
```
Vid201, line 21: We should split `thirdparty/account-abstraction` and `thirdparty/bundler` and have the latter only on `make test`. I do not think any of the sc will be used in the main build.

.github/workflows/ci.yml
```diff
@@ -19,6 +19,10 @@ jobs:
       - name: Setup third party dependencies
         run: |
           make fetch-thirdparty
+      - name: Install Foundry
```
Vid201, line 22: I saw you implemented `debug_trace_call` in Foundry. We can add to README.md that you can use Foundry during testing/dev.

Vid201: And info about Foundry to README.md.

tests/validate_tests.rs
```diff
@@ -0,0 +1,512 @@
...
+use ethers::types::Address;
+use ethers::utils::{parse_units, AnvilInstance};
+use ethers::{
+    core::utils::Anvil,
+    prelude::SignerMiddleware,
+    providers::Provider,
+    signers::{LocalWallet, Signer},
+    types::{Bytes, U256},
+};
+use std::ops::Deref;
```
Vid201, line 22: unused import?

tests/validate_tests.rs
```diff
@@ -0,0 +1,512 @@
...
+    test_existing_user_op("balance-self".to_string(), "".to_string())
+        .await
+        .expect("succeed");
+}
+
+#[tokio::test]
+#[should_panic]
+async fn fail_with_unstaked_paymaster_returning_context() {
+    let context = get_global_context().await;
+    let pm = deploy_test_storage_account(context.client.clone())
```
Vid201, line 429: I get error: 'type annotations needed cannot infer type'.

tests/validate_tests.rs
```diff
@@ -0,0 +1,512 @@
...
+        init_func,
+        context.storage_factory.address,
+    )
+    .await
+    .expect("succeed");
+}
+
+#[tokio::test]
+#[should_panic]
+async fn fail_with_inner_oog_revert() {
```
Vid201, line 498: this test fails: 'test did not panic as expected'

Vid201 (changes requested):

Vid201:
Looks good @zsluedem. I wonder if these tests will stay relevant in the future or if we should focus on https://github.com/eth-infinitism/bundler-spec-tests. This was released two weeks ago and will be the official test suite for bundlers (maybe integrate running these tests instead in the repo).

What do you think?

zsluedem:
> Looks good @zsluedem. I wonder if these tests will stay relevant in the future or if we should focus on https://github.com/eth-infinitism/bundler-spec-tests. This was released two weeks ago and will be the official test suite for bundlers (maybe integrate running these tests instead in the repo).
> 
> What do you think?

Hm...I didn't know the test suite. I read the test codes there today and compare. They got very similar test contracts for opcode-banning parts.

The test suite would be run with a bundler client and an execution client from higher-layer integration perspective. I think we could have both because this could help you develop the simulation checking function (I think). You don't need to have a fully functional bundler to run the test suite right now.

But I am Ok if you want to just use the test suite.

tests/validate_tests.rs
zsluedem: It is used in https://github.com/Vid201/aa-bundler/blob/4d55d1d3ea52d2780ad0cdd0c1d79af3758c1ccc/tests/validate_tests.rs#L112.
Currently, clippy should be enough for unused import.

tests/validate_tests.rs
zsluedem: There are some underministic problem about the tests. I think it is caused by the singleton of the TestContext. I am looking at it.

tests/validate_tests.rs
Vid201: Yes, it's better now. Nice catch to add the version to Cargo.toml.

README.md
```diff
@@ -13,7 +13,8 @@ For more information: https://hackmd.io/@Vid201/aa-bundler-rust
 ## Prerequisites
 
 1. Ethereum JSON-RPC API with enabled [`debug_traceCall`](https://geth.ethereum.org/docs/interacting-with-geth/rpc/ns-debug#debug_tracecall) (currently implemented only in [Geth](https://github.com/ethereum/go-ethereum) and [Erigon](https://github.com/ledgerwatch/erigon)). For testing purposes, you can setup [private Geth node](https://github.com/krzkaczor/geth-private-node).  
-1. [solc](https://docs.soliditylang.org/en/v0.8.17/installing-solidity.html) >=0.8.12
+2. [solc](https://docs.soliditylang.org/en/v0.8.17/installing-solidity.html) >=0.8.12
+3. For testing purposes, we need [Anvil](https://github.com/foundry-rs/foundry/tree/master/anvil#anvil) which would provide enough functions of execution client like Geth. To install Anvil, you can setup with [foundryup](https://getfoundry.sh/)
```
Vid201, line 17: I would combine 1. and 3. Something like that:

1. Ethereum execution client JSON-RPC API with enabled [`debug_traceCall`](https://geth.ethereum.org/docs/interacting-with-geth/rpc/ns-debug#debug_tracecall). For production, you can use [Geth](https://github.com/ethereum/go-ethereum) or [Erigon](https://github.com/ledgerwatch/erigon). For testing purposes, you can use [Anvil](https://github.com/foundry-rs/foundry/tree/master/anvil#anvil), which provides enough functionalities of execution clients like Geth. To install Anvil, use [foundryup](https://getfoundry.sh/)

zsluedem: Cool. Love your version.

Vid201:
LGTM :rocket:

Vid201 (approved):

PR #31, opened 2023-01-02
Fix CI failure

zsluedem:


.github/workflows/CI.yml
```diff
@@ -1,10 +1,14 @@
...
+    branches:
+      - main
+  pull_request:
 
 name: CI
 
 jobs:
   build_and_test:
-    name: Rust project
+    name: Rust Abstract Account Project
```
Vid201, line 11: Account Abstraction Bundler Rust

Vid201 (changes requested):

.github/workflows/CI.yml
```diff
@@ -15,6 +19,15 @@ jobs:
...
           make fetch-thirdparty
+      - name: Install solc
+        run: |
+          sudo add-apt-repository ppa:ethereum/ethereum
+          sudo add-apt-repository ppa:ethereum/ethereum-dev
+          sudo apt-get update
+          sudo apt-get install solc
+      - name: Build
+        run: |
+          cargo build
```
Vid201, line 30: Let's move this `make build` ... new entry to Makefile, so we have all commands in single place

Vid201 (approved):

PR #34, opened 2023-01-02
Memory user operation pool

Vid201:


src/utils.rs
```diff
@@ -0,0 +1,9 @@
+use ethers::types::{Address, U256};
+use std::str::FromStr;
+
+pub fn parse_address(s: &str) -> Result<Address, String> {
+    Address::from_str(s).map_err(|_| format!("Adress {} is not a valid address", s))
+}
+pub fn parse_u256(s: &str) -> Result<U256, String> {
+    U256::from_str_radix(s, 10).map_err(|_| format!("{} is not a valid U256", s))
+}
```
zsluedem, line 9: I think you could create a small pr for this kind of change which could cause less conflicts for your codes later

src/uopool/memory.rs
```diff
@@ -0,0 +1,352 @@
...
+            .read()
+            .contains_key(&user_operation_hash)
+        {
+            return Err(anyhow::anyhow!("User operation not found"));
+        }
+        Ok(self
+            .user_operations
+            .read()
+            .get(&user_operation_hash)
+            .unwrap()
```
zsluedem, line 82: Unsafe unwrap here. This function return `Result`. It would be better turns it into Err

src/uopool/memory.rs
```diff
@@ -0,0 +1,352 @@
...
+        &self,
+        entry_point: Address,
+        chain_id: U256,
+    ) -> anyhow::Result<Vec<UserOperation>> {
+        let id = MemoryMempool::id(entry_point, chain_id);
+        Ok(self
+            .user_operations_by_entry_point
+            .read()
+            .get(&id)
+            .unwrap()
```
zsluedem, line 100: unsafe unwrap

src/uopool/memory.rs
```diff
@@ -0,0 +1,352 @@
...
+        sender: Address,
+        entry_point: Address,
+        chain_id: U256,
+    ) -> anyhow::Result<Vec<UserOperation>> {
+        let id = MemoryMempool::id(entry_point, chain_id);
+        Ok(self
+            .user_operations_by_sender
+            .read()
+            .get(&id)
+            .unwrap()
```
zsluedem, line 117: unsafe unwrap

src/uopool/mod.rs
```diff
@@ -1,36 +1,65 @@
...
+pub type MempoolId = H256;
 
-impl UserOperationPool {
-    pub fn new() -> Self {
-        Self {
-            pool: Default::default(),
-        }
+#[async_trait]
+pub trait Mempool: Debug + Send + Sync + 'static {
+    fn id(entry_point: Address, chain_id: U256) -> MempoolId {
```
zsluedem, line 28: I think it would be better to make the `id` into a separate trait. Just a personal preference.

src/uopool/memory.rs
```diff
@@ -0,0 +1,352 @@
...
+use super::{Mempool, MempoolId};
+
+pub type UserOperationsBySender = HashMap<Address, HashSet<UserOperationHash>>;
+
+#[derive(Educe)]
+#[educe(Debug)]
+pub struct MemoryMempool {
+    user_operations: Arc<RwLock<HashMap<UserOperationHash, UserOperation>>>, // user_operation_hash -> user_operation
+    user_operations_by_entry_point: Arc<RwLock<HashMap<MempoolId, HashSet<UserOperationHash>>>>, // mempool_id -> user_operations
+    user_operations_by_sender: Arc<RwLock<HashMap<MempoolId, UserOperationsBySender>>>, // mempool_id -> sender -> user_operations
```
zsluedem, line 20: I think in the current stage it is kind of over-designed to have several entry_point mempools.

I was thinking that it is more flexible to have a simple `HashMap<UserOperationHash, UserOperation>` mempool and it could be extended to multiple entrypoint mempools by wrapping the simple mempool.

src/uopool/mod.rs
```diff
@@ -1,36 +1,65 @@
...
+        H256::from_slice(keccak256([entry_point.encode(), chain_id.encode()].concat()).as_slice())
     }
+    fn add(
```
zsluedem, line 31: I am curious. Why not make these methods in `async` way? In that way, we could support cocurrent.

zsluedem, line 31: I am curious. Why not make these methods in `async` way? In that way, we could support cocurrent.

zsluedem (changes requested):
Please try to avoid unsafe `unwrap`.

src/uopool/memory.rs
```diff
@@ -0,0 +1,352 @@
...
+            let user_operation_old = user_operations.get(&hash).unwrap();
+            let max_priority_fee_per_gas_diff = user_operation.max_priority_fee_per_gas
+                - user_operation_old.max_priority_fee_per_gas;
+            let max_fee_per_gas_diff =
+                user_operation.max_fee_per_gas - user_operation_old.max_fee_per_gas;
+            if !(user_operation.sender == user_operation_old.sender
+                && user_operation.nonce == user_operation_old.nonce
+                && user_operation.max_priority_fee_per_gas
+                    > user_operation_old.max_priority_fee_per_gas
+                && max_priority_fee_per_gas_diff == max_fee_per_gas_diff)
```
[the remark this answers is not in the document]
Vid201: Hmm, I checked the specs again, and I misunderstood them.

I will move this check (with some changes; you're right about the hash) to the sanity checks in other PR.

src/uopool/mod.rs
Vid201: Hmm, missed that.

src/uopool/memory.rs
Vid201: What about HashMap at the `UoPoolService` level:
`pools: ... HashMap<MempoolId, Mempool>`
?

src/uopool/mod.rs
Vid201: We can do that, yeah.

src/uopool/mod.rs
zsluedem: Sorry, @Vid201 .  I was also thinking of replacing the Rwlock using the async-typed lock to gain better concurrent performance. But the truth is it won't provide better performance with the async-typed lock from this [doc](https://docs.rs/tokio/latest/tokio/sync/struct.Mutex.html#which-kind-of-mutex-should-you-use). So it might be better to stick with your original solutions.

I am sorry that I mislead you.

src/uopool/memory.rs
zsluedem: That sounds better.

src/uopool/mod.rs
zsluedem: I am not quite sure which is the best solution right now. I am OK with either sync or async function right now.

src/uopool/mod.rs
```diff
@@ -1,36 +1,52 @@
...
+pub trait Mempool: Debug + Send + Sync + 'static {
+    async fn add(
+        &mut self,
+        user_operation: UserOperation,
+        entry_point: Address,
+        chain_id: U256,
+    ) -> anyhow::Result<UserOperationHash>;
+    async fn get(&self, user_operation_hash: UserOperationHash) -> anyhow::Result<UserOperation>;
+    async fn get_all(&self) -> anyhow::Result<Vec<UserOperation>>;
+    async fn get_all_by_sender(&self, sender: Address) -> anyhow::Result<Vec<UserOperation>>;
```
zsluedem, line 41: Just a nitpick. I think we could also turn the `get_all` and `get_all_by_sender` into an `Iterator` trait method which could simplify the trait even more.

src/types/user_operation.rs
```diff
@@ -74,6 +74,22 @@ impl UserOperation {
             .as_slice(),
         )
     }
+
+    pub fn random() -> Self {
```
zsluedem, line 78: mark this method into `#[cfg(test)]`?

zsluedem (commented):
Great~! I would probably add [additional lint for `unwrap`](https://rust-lang.github.io/rust-clippy/master/index.html#unwrap_used) in another pr.

src/uopool/mod.rs
Vid201: No problem @zsluedem . I have some experience with concurrency in Rust, but still learning a lot as we go.

I think it's okay if we stick with the current solution (async fn); let's see if this needs to be changed later.

zsluedem (approved):
Thanks for addressing all the comments.

PR #35, opened 2023-01-03
Misc updates

Vid201:
This PR contains several small updates.


bin/bundler.rs
```diff
@@ -18,6 +20,9 @@ pub struct Opt {
     #[clap(long)]
     pub mnemonic_file: ExpandedPathBuf,
 
+    #[clap(long, value_delimiter=',', value_parser=parse_address)]
+    pub entry_points: Vec<Address>,
```
zsluedem, line 24: I am just curious what kind of situations are you thinking about multiple entry_point contract. There were discussion in the discord that dror said that entry_point should be a singleton (https://discord.com/channels/892780451570270219/892780453940056066/1056914361047855114).

zsluedem (approved):
LGTM. Left a question.

bin/bundler.rs
Vid201: Hmm, interesting. 

There is still this endpoint in the EIP: https://github.com/eth-infinitism/account-abstraction/blob/develop/eip/EIPS/eip-4337.md#-eth_supportedentrypoints. Meaning multiple entry points can be supported by the bundler.

Upgrading the entry point smart contract could also lead to multiple entry points most likely: https://github.com/eth-infinitism/account-abstraction/blob/develop/eip/EIPS/eip-4337.md#entry-point-upgrading.

zsluedem: Good point.

PR #38, opened 2023-01-14
Add unwrap check to lint

zsluedem:


src/types/user_operation.rs
```diff
@@ -45,20 +52,17 @@ impl From<UserOperation> for entry_point_api::UserOperation {
...
-        );
-        encoded.push_str(&packed[..packed.len() - 64]);
-        encoded.parse::<Bytes>().unwrap()
+        let mut packed: Vec<u8> = UserOperation {
+            signature: Bytes::default(),
+            ..self.clone()
+        }
+        .encode();
+        packed.truncate(packed.len() - 32);
+        Bytes::from(packed)
```
zsluedem, line 65: We don't have to convert from bytes to str and then str to bytes which is unnecessary and slower. We could just operate on vec bytes.

src/types/user_operation.rs
```diff
@@ -5,25 +5,32 @@ use ethers::types::{Address, Bytes, TransactionReceipt, H256, U256};
...
-    pub call_gas_limit: U256,
-    pub verification_gas_limit: U256,
-    pub pre_verification_gas: U256,
-    pub max_fee_per_gas: U256,
-    pub max_priority_fee_per_gas: U256,
-    pub paymaster_and_data: Bytes,
-    pub signature: Bytes,
+// In order to mute unwrap used lint error which is generated by the #derive(EthAbiType)
+// UserOperation is put in the private mod and mute the error.
+#[allow(clippy::unwrap_used)]
```
zsluedem, line 13: EthAbiType would generate unwrap which I could not avoid now.

zsluedem: https://github.com/gakonst/ethers-rs/issues/2053 created an issue for ethers-rs

Vid201: @zsluedem Do you want to wait with this PR to see if this fix in `ethers-rs` will be merged in the coming days (we can then just update the lib version)?

src/types/user_operation.rs
Vid201: Looks good

src/types/user_operation.rs
zsluedem: Hm.. I am fine with either way.

Vid201: Then let's wait.

zsluedem: @Vid201 Updated now.

Vid201 (approved):

PR #39, opened 2023-01-15
RPC endpoint eth_sendUserOperation

Vid201:
Implementation of `eth_sendUserOperation`.

Next steps are to run sanity checks and simulations in the `uopool` component.

### DIFF
--- a/src/uopool/server.rs
+++ b/src/uopool/server.rs
@@ -88,7 +88,7 @@ pub mod types {
             }
         }
     }
-    
+
     impl From<UserOperation> for crate::types::user_operation::UserOperation {
         fn from(user_operation: UserOperation) -> Self {
             Self {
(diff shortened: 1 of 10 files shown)


src/uopool/services/uopool.rs
```diff
@@ -30,12 +33,38 @@ impl UoPoolService {
...
+                Some(json!({
+                    "reason": "this is error",
+                })),
+            );
+
+            res.set_result(AddResult::NotAdded);
+            res.data = serde_json::to_string(&uo_pool_error)
+                .map_err(|_| tonic::Status::internal("error adding user operation"))?;
+
+            return Ok(tonic::Response::new(res));
```
zsluedem, line 64: Although I got questions with the codes above, I think because it is not complete yet.

src/uopool/server.rs
```diff
@@ -1,7 +1,119 @@
-mod types {
+// Code adapted from: https://github.com/ledgerwatch/interfaces/blob/master/src/lib.rs#L1
+pub mod types {
```
zsluedem, line 2: good work

zsluedem (approved):
Great work!

src/uopool/services/uopool.rs
Vid201: Yes, that's right, this part is not yet completely done.

PR #43, opened 2023-01-22
Reputation

Vid201:
Several sanity checks and simulation conditions are based on the reputation of entities (paymasters, senders, aggregators, factories). Thus we should have a reputation implementation first.

This PR contains:

- reputation management
- debug endpoints
- small changes to the Makefile


bin/bundler.rs
```diff
@@ -87,6 +90,13 @@ fn main() -> Result<()> {
                             ))
                             .await?;
 
+                            #[cfg(debug_assertions)]
```
zsluedem, line 93: Is this flag necessary? I think we would still need this in production build.

src/uopool/mod.rs
```diff
@@ -18,18 +25,20 @@ use jsonrpsee::tracing::info;
...
-pub mod memory;
+pub mod memory_mempool;
+pub mod memory_reputation;
 pub mod server;
 pub mod services;
 
 pub type MempoolId = H256;
 
+pub type MempoolBox<T> = Box<dyn Mempool<UserOperations = T>>;
+pub type ReputationBox<T> = Box<dyn Reputation<ReputationEntries = T>>;
```
zsluedem, line 36: I am curious why this needs to be `dyn` now. Why not just use `MemoryMempool` and `MemoryReputation`.

zsluedem (commented):
Good work! I left some comments and questions.

bin/bundler-rpc.rs
```diff
@@ -34,6 +37,14 @@ async fn main() -> Result<()> {
     let mut api = Methods::new();
     let uopool_grpc_client =
         UoPoolClient::connect(format!("http://{}", opt.uopool_grpc_listen_address)).await?;
+
+    #[cfg(debug_assertions)]
```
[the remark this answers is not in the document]
Vid201: Debug API should be available only for passing the bundler spec tests. In production release, they must be turned off.

src/uopool/mod.rs
```diff
@@ -53,24 +102,57 @@ pub struct UserOperationPool<M: Mempool> {
...
             mempools.insert(id, Box::<MemoryMempool>::default());
+
+            reputations.insert(id, Box::<MemoryReputation>::default());
+            if let Some(reputation) = reputations.get_mut(&id) {
+                reputation.init(
+                    MIN_INCLUSION_RATE_DENOMINATOR,
+                    THROTTLING_SLACK,
+                    BAN_SLACK,
+                    opts.min_stake,
+                    opts.min_unstake_delay,
```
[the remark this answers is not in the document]
Vid201: Yes, the first three are config parameters that are the same between all different entry points.

src/uopool/mod.rs
Vid201: Well, my idea was that sometimes, someone would want to use `MemoryMempool` for one mempool and `DatabaseMempool` for another. But at the moment, this is somewhat complex, yeah.

zsluedem: Hm. Ok. Let's leave it as it is now.

src/uopool/mod.rs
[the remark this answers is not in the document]
zsluedem: Cools

zsluedem (approved):
Nice work.

PR #48, opened 2023-02-04
chore: remove async

Vid201:
There is a `RwLock` for mempools and reputations on the UoPoolService level (`Arc<RwLock<HashMap<MempoolId...`), and both will always be accessed over these two hashmaps, thus we don't need async traits for Mempool and Reputation.

### DIFF
--- a/src/uopool/memory_mempool.rs
+++ b/src/uopool/memory_mempool.rs
@@ -1,11 +1,7 @@
 use async_trait::async_trait;
 use educe::Educe;
 use ethers::types::{Address, U256};
-use parking_lot::RwLock;
-use std::{
-    collections::{HashMap, HashSet},
-    sync::Arc,
-};
+use std::collections::{HashMap, HashSet};
 
 use crate::types::user_operation::{UserOperation, UserOperationHash};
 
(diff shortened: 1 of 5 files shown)


src/uopool/memory_mempool.rs
```diff
@@ -14,75 +10,70 @@ use super::Mempool;
 #[derive(Default, Educe)]
 #[educe(Debug)]
 pub struct MemoryMempool {
-    user_operations: Arc<RwLock<HashMap<UserOperationHash, UserOperation>>>, // user_operation_hash -> user_operation
-    user_operations_by_sender: Arc<RwLock<HashMap<Address, HashSet<UserOperationHash>>>>, // sender -> user_operations
+    user_operations: HashMap<UserOperationHash, UserOperation>, // user_operation_hash -> user_operation
+    user_operations_by_sender: HashMap<Address, HashSet<UserOperationHash>>, // sender -> user_operations
```
zsluedem, line 14: I am curious what is the intention to remove the `RwLock` primitives here?

zsluedem (commented):
Good work. I want to remove the async in these traits, too. Glad that we want to do the same. I left one question about the primitives`RwLock`

src/uopool/memory_mempool.rs
Vid201: I don't think they are needed and bring some additional overhead. We have another `RwLock` here: https://github.com/Vid201/aa-bundler/blob/main/src/uopool/services/uopool.rs#L25. Mempools are always accessed through that `RwLock`, so the whole mempool should be anyway thread-safe.

What do you think?

zsluedem: Oh. That makes sense. If there are some performance problem in the future, we could make lock scope smaller at that time. Cool.

zsluedem (approved):
Great

src/uopool/memory_mempool.rs
Vid201: Exactly, great!

PR #49, opened 2023-02-06
Introduce mdbx UserOperation Pool

zsluedem:
Trying to address https://github.com/Vid201/aa-bundler/issues/8.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,14 +15,18 @@ clap = { version = "4", features = ["derive"] }
 dirs = "4.0"
 educe = { version = "0.4", features = ["Debug", "Default"] }
 ethereum-interfaces = { git = "https://github.com/ledgerwatch/interfaces" }
-ethers = { git = "https://github.com/gakonst/ethers-rs.git", rev = "c60990d7b6e6b3034f1e50b27b495aeda30e35e8", features = ["ethers-solc"]  }
+ethers = { git = "https://github.com/gakonst/ethers-rs.git", rev = "c60990d7b6e6b3034f1e50b27b495aeda30e35e8", features = ["ethers-solc"] }
 expanded-pathbuf = "0.1"
 hex = { version = "0.4.3", default-features = false, features = ["std"] }
 jsonrpsee = { version = "0.16", features = ["server", "macros"] }
+page_size = "0.5.0"
 parking_lot = "0.12"
 prost = "0.11"
 regex = "1"
+reth-db = { git = "https://github.com/paradigmxyz/reth.git", rev = "314ea0883b1062375ad86577be1fccc39233739d" }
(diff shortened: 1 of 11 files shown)


Cargo.toml
```diff
@@ -33,6 +33,10 @@ tonic = { version = "0.8", default-features = false, features = [
 ] }
 tracing = "0.1"
 tracing-subscriber = "0.3"
+rustc-hex="^2.0.1"
```
Vid201, line 36: sort deps by alphabet

src/types/mod.rs
```diff
@@ -1,2 +1,33 @@
+use ethers::types::Address;
+use reth_db::table::{Decode, Encode};
+
 pub mod reputation;
 pub mod user_operation;
+
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone)]
+pub struct WrapAddress(Address);
```
Vid201, line 8: I would move this from `mod.rs` to separate file, e.g., `src/types/utils.rs` or something like that.

All WrapAddress things.

src/uopool/database.rs
```diff
@@ -0,0 +1,328 @@
...
+
+    os_page_size.clamp(min_page_size, libmdbx_max_page_size)
+}
+
+impl<E: EnvironmentKind> UserOpDatabase<E> {
+    pub fn new(path: PathBuf) -> anyhow::Result<Self> {
+        let env = Environment::new()
+            .set_max_dbs(TABLES.len())
+            .set_geometry(Geometry {
+                size: Some(0..(1024 * 1024 * 1024 * 1024 * 4)), // TODO: reevaluate (4 tb)
```
Vid201, line 192: Are there any recommendations somewhere how to set these values?

Vid201 (changes requested):

Vid201:
This looks really good.

Good job!

Cargo.toml
zsluedem: https://crates.io/crates/cargo-sort introduce cargo sort in ci lint  check.

Vid201: Can you add this in this PR? It's not a big thing.

src/uopool/database.rs
zsluedem: Some doc here https://erthink.github.io/libmdbx/group__c__settings.html#ga79065e4f3c5fb2ad37a52b59224d583e.

Based on my previous experiences on lmdb, the size here is upper bound of disk consumption and it depends on the available disk space of the machine. I saw many apps setting this value very large because the lib would throw errors if the size of the database reach to the value.

Cargo.toml
zsluedem: Added
b94b2820e9552f00064f5b83dedaf0af7f633683

Vid201 (approved):

PR #52, opened 2023-02-11
feat: add docker support and bundler-spec-test startup

Vid201:
The easiest way to run bundler spec test (https://github.com/eth-infinitism/bundler-test-executor) is with Docker.

This PR also works towards #18.


Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+
+COPY --from=builder /aa-bundler/target/debug/create-wallet /usr/local/bin/create-wallet
+COPY --from=builder /aa-bundler/target/debug/bundler /usr/local/bin/bundler
+
+# $HOME == /root
+RUN if [ -z "$(ls -A $HOME)" ]; then /usr/local/bin/create-wallet --output-path $HOME ; fi
+
+EXPOSE 3000
+
+ENTRYPOINT usr/local/bin/bundler --rpc-listen-address 0.0.0.0:3000 --eth-client-address http://geth-dev:8545 --mnemonic-file $HOME/$(ls $HOME | head -1) --beneficiary 0x690B9A9E9aa1C9dB991C7721a92d351Db4FaC990 --gas-factor 600 --min-balance 1 --entry-points 0x0000000000000000000000000000000000000000 --chain-id 5 --helper 0x0000000000000000000000000000000000000000 --min-stake 1 --min-unstake-delay 0
```
zsluedem, line 36: Why do we need to config all the parameters here? It should be configured when running the container.

zsluedem (approved):
Looks great

PR #59, opened 2023-02-16
eth_getSupportedEntryPoints and eth_estimateUserOperationGas

Vid201:


src/types/user_operation.rs
```diff
@@ -155,6 +156,100 @@ pub struct UserOperationReceipt {
     pub receipt: TransactionReceipt,
 }
 
+#[derive(Serialize, Deserialize)]
+pub struct UserOperationPartial {
```
Vid201, line 160: @zsluedem Do you know of any better approach?

src/proto/uopool/uopool.proto
```diff
@@ -44,6 +44,33 @@ message GetChainIdResponse {
...
 
+message GetSupportedEntryPointsRequest {}
+
+enum GetSupportedEntryPointsResult {
+    GOT_SUPPORTED_ENTRY_POINTS = 0;
+    NOT_GOT_SUPPORTED_ENTRY_POINTS = 1;
+}
+
+message GetSupportedEntryPointsResponse {
+    GetSupportedEntryPointsResult result = 1;
```
zsluedem, line 55: In practice, it would always `GOT_SUPPORTED_ENTRY_POINTS`. Is it neccessary to add this?

src/types/user_operation.rs
zsluedem: No, I don't.

zsluedem (changes requested):
It has conflicts now.

src/proto/uopool/uopool.proto
```diff
@@ -100,11 +110,13 @@ message SetReputationResponse {
 service UoPool {
     rpc Add(AddRequest) returns (AddResponse);
     rpc Remove(RemoveRequest) returns (RemoveResponse);
-    rpc GetChainId(GetChainIdRequest) returns (GetChainIdResponse);
+    rpc GetChainId(google.protobuf.Empty) returns (GetChainIdResponse);
```
zsluedem, line 113: Good

zsluedem (approved):
LGTM

PR #61, opened 2023-02-17
fix: update entry point simulate validation trace

Vid201:
Account abstraction (third party) commit now points to the latest released version: v0.4.0.


zsluedem (approved):
Great.

PR #62, opened 2023-02-20
First iteration of simulate validation (user operation)

Vid201:
This adds first checks for simulate validation - forbidden opcodes. It also removes `sanity_check_results` and rertieves the stake info at the start (alongside JS trace).

### DIFF
--- a/src/uopool/services/sanity_check.rs
+++ b/src/uopool/services/sanity_check.rs
@@ -3,7 +3,7 @@ use crate::{
     contracts::EntryPointErr,
     types::{
         reputation::{ReputationStatus, StakeInfo},
-        sanity_check::{BadUserOperationError, SanityCheckResult},
+        sanity_check::BadUserOperationError,
         user_operation::UserOperation,
     },
     uopool::{mempool_id, services::uopool::UoPoolService},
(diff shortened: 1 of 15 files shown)


src/uopool/services/sanity_check.rs
```diff
@@ -296,17 +240,9 @@ where
...
-        self.sanity_check_results.write().insert(
-            user_operation.hash(entry_point, &self.chain_id),
-            Default::default(),
-        );
-
         // Either the sender is an existing contract, or the initCode is not empty (but not both)
         self.sender_or_init_code(user_operation).await?;
 
-        // If initCode is not empty, parse its first 20 bytes as a factory address. Record whether the factory is staked, in case the later simulation indicates that it needs to be. If the factory accesses global state, it must be staked - see reputation, throttling and banning section for details.
-        self.verify_factory(user_operation, entry_point).await?;
```
zsluedem, line 308: Curious why this is removed.

zsluedem (commented):
I am curious what stuffs left to be resolved

Vid201:
> I am curious what stuffs left to be resolved

You mean in general or in the simulation? For simulation, there are still several things: storage access, limitations on CALL opcodes ... I will continue with these things.

src/uopool/services/sanity_check.rs
Vid201: The only thing that was verified here is if the factory is staked. This is now available here: https://github.com/Vid201/aa-bundler/blob/feat/uo_trace/src/uopool/services/simulation.rs#L98, so no need to check that here also.

zsluedem (approved):
LGTM

PR #65, opened 2023-02-28
chore: update deps

Vid201:
@zsluedem Some tests in validate_tests.rs are failing, if you could take a look (maybe it's related to some updates in smart contracts?).

### DIFF
--- a/tests/validate_tests.rs
+++ b/tests/validate_tests.rs
@@ -218,6 +218,7 @@ async fn test_existing_user_op(validate_rule: String, pm_rule: String) -> anyhow
 }
 
 #[tokio::test]
+#[ignore]
 async fn accept_plain_request() -> anyhow::Result<()> {
     let context = setup().await?;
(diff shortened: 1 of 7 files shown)


tests/validate_tests.rs
```diff
@@ -218,6 +218,7 @@ async fn test_existing_user_op(validate_rule: String, pm_rule: String) -> anyhow
 }
 
 #[tokio::test]
+#[ignore]
 async fn accept_plain_request() -> anyhow::Result<()> {
```
zsluedem, line 222: These tests are not actually working right now because the validate function hasn't implement in https://github.com/Vid201/aa-bundler/blob/398d24623db5021906c556de9daede998af5f051/tests/validate_tests.rs#L192-L195.

I created an issue for it https://github.com/Vid201/aa-bundler/issues/66.

zsluedem (approved):
LGTM

PR #71, opened 2023-03-09
Feat: storage access verification

Vid201:


src/uopool/services/simulation.rs
```diff
@@ -151,9 +330,23 @@ where
             }
         })?;
 
+        let mut stake_info_by_entity: HashMap<usize, StakeInfo> = HashMap::new();
```
zsluedem, line 333: I think fixed array `[StakeInfo;3]` is a better data structure here.

zsluedem: Is there a good reason to use HashMap here?

src/uopool/services/simulation.rs
```diff
@@ -86,46 +91,220 @@ where
...
+    }
+
+    fn storage_access(
+        &self,
+        user_operation: &UserOperation,
+        entry_point: &Address,
+        stake_info_by_entity: &HashMap<usize, StakeInfo>,
+        trace: &JsTracerFrame,
+    ) -> Result<(), SimulateValidationError<M>> {
+        let mut slots_by_entity = HashMap::new();
```
zsluedem, line 255: Is it possible to use just `HashMap<Address, HashSet<Bytes>>` instead of `HashMap<Address, HashSet<String>>` which we could skip some bytes to string and string to bytes converts? I am just curious.

zsluedem (commented):
Left some questions.

src/uopool/services/simulation.rs
Vid201: Yes, you're right. Will change that.

src/uopool/services/simulation.rs
Vid201: Nice catch, that makes more sense.

Vid201:
@zsluedem

zsluedem (approved):
Very good!

PR #73, opened 2023-03-12
Enable debug API through cmd flag, build docker image with CI

Vid201:
There is a new CI: it publishes a docker image to ghcr on each push to main branch.


zsluedem (approved):
Nice job.

PR #82, opened 2023-03-23
Remove JsonRpcError

zsluedem:
After upgrading ethers-rs in https://github.com/Vid201/aa-bundler/pull/79.

The JsonRpcError is not reliadble source of revert error because the api changed in ethers-rs.

This pr is intended to remove JsonRpcError struct.

### DIFF
--- a/src/uopool/services/uopool.rs
+++ b/src/uopool/services/uopool.rs
@@ -220,22 +220,14 @@ where
                             }
                             Err(error) => {
                                 res.set_result(EstimateUserOperationGasResult::NotEstimated);
-                                res.data =
-                                    serde_json::to_string(&SimulationError::from(match error {
-                                        EntryPointErr::JsonRpcError(err) => {
-                                            SimulateValidationError::<M>::UserOperationExecution {
-                                                message: err.message,
-                                            }
-                                        }
-                                        _ => SimulateValidationError::UnknownError {
-                                            error: format!("{error:?}"),
-                                        },
-                                    }))
-                                    .map_err(|_| {
-                                        tonic::Status::internal(
-                                            "error estimating user operation gas",
-                                        )
-                                    })?;
+                                res.data = serde_json::to_string(&SimulationError::from(
+                                    SimulateValidationError::<M>::UnknownError {
+                                        error: format!("{error:?}"),
+                                    },
+                                ))
+                                .map_err(|_| {
+                                    tonic::Status::internal("error estimating user operation gas")
+                                })?;
                             }
                         }
                     }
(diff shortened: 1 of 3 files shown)


src/types/simulation.rs
```diff
@@ -55,7 +54,6 @@ pub struct StakeInfo {
 pub enum SimulateValidationError<M: Middleware> {
     UserOperationRejected { message: String },
     OpcodeValidation { entity: String, opcode: String },
-    UserOperationExecution { message: String },
```
zsluedem, line 58: @Vid201 I don't think this is needed now. But I need your confirmation on it.

src/types/sanity_check.rs
```diff
@@ -48,9 +47,6 @@ pub enum BadUserOperationError<M: Middleware> {
     SenderVerification {
         sender: Address,
     },
-    UserOperationExecution {
-        message: String,
-    },
```
zsluedem, line 53: @Vid201 I don't think this is needed now. But I need your confirmation on it.

src/types/simulation.rs
Vid201: It is needed for this test: https://github.com/eth-infinitism/bundler-spec-tests/blob/master/tests/rpc/test_eth_estimateUserOperationGas.py#L23.

I suggest to setup local environment for running tests, so that these changes won't break the bundler spec tests (https://github.com/eth-infinitism/bundler-spec-tests).

Vid201: Just start local geth, run bundler (`run-bundler-debug`), and deploy entry point. Then you can run tests from bundler-spec-tests repo: `pdm run pytest -rA -W ignore::DeprecationWarning --url http://127.0.0.1:3000/ --entry-point 0x0576a174D229E3cFA37253523E645A78A0C91B57 --ethereum-node http://127.0.0.1:8545/`

Vid201:
Yeah, you're right. I just run the executor and the tests are failing.

src/uopool/services/uopool.rs
```diff
@@ -220,22 +220,14 @@ where
...
-                                            error: format!("{error:?}"),
-                                        },
-                                    }))
-                                    .map_err(|_| {
-                                        tonic::Status::internal(
-                                            "error estimating user operation gas",
-                                        )
-                                    })?;
+                                res.data = serde_json::to_string(&SimulationError::from(
+                                    SimulateValidationError::<M>::UnknownError {
```
Vid201, line 224: This should probably be handled differently, the UserOperationExecution is never returned.

Vid201 (approved):

Vid201:
Check #84

src/types/simulation.rs
zsluedem: Makes sense. I went through the official spec and didn't find it. It turns out to be in bundler-spec-tests.

PR #85, opened 2023-03-25
Fix execution revert case in estimate gas api

zsluedem:
Fix https://github.com/Vid201/aa-bundler/issues/84

Middleware error should be all internal errors which bundler user should not know that. I remove the trait bound on the error type. Be more concrete on the error.

### DIFF
--- a/src/types/simulation.rs
+++ b/src/types/simulation.rs
@@ -1,5 +1,5 @@
 use ethers::types::{Address, U256};
-use jsonrpsee::types::ErrorObject;
+use jsonrpsee::types::{error::ErrorCode, ErrorObject};
 use lazy_static::lazy_static;
 use std::collections::HashSet;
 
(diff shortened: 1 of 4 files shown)


src/types/simulation.rs
```diff
@@ -84,11 +80,8 @@ impl<M: Middleware> From<SimulateValidationError<M>> for SimulationError {
             SimulateValidationError::CallStackValidation { message } => {
                 SimulationError::owned(OPCODE_VALIDATION_ERROR_CODE, message, None::<bool>)
             }
-            SimulateValidationError::Middleware(_) => {
-                SimulationError::from(ErrorCode::InternalError)
-            }
             SimulateValidationError::UnknownError { error } => {
-                SimulationError::owned(SIMULATE_VALIDATION_ERROR_CODE, error, None::<bool>)
+                SimulationError::owned(ErrorCode::InternalError.code(), error, None::<bool>)
```
zsluedem, line 84: @Vid201 I am curious why unknow error belongs to `SIMULATE_VALIDATION_ERROR_CODE`.

zsluedem: I changed that into internal error. But like your ideas on it.

Vid201 (approved):

Vid201:
LGTM!

PR #87, opened 2023-03-26
Feat/codehash

Vid201:
Check that code hashes of accessed addresses didn't changed between 1st and 2nd simulation.

### DIFF
--- a/src/uopool/services/simulation.rs
+++ b/src/uopool/services/simulation.rs
@@ -415,7 +415,7 @@ impl<M: Middleware + 'static> UoPoolService<M> {
             let eth_provider = self.eth_provider.clone();
 
             tasks.spawn(async move {
-                match eth_provider.get_code(contract_address.clone(), None).await {
+                match eth_provider.get_code(contract_address, None).await {
                     Ok(code) => Some((contract_address, keccak256(&code).into())),
                     Err(_) => None,
                 }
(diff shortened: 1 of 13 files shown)


src/uopool/services/simulation.rs
```diff
@@ -403,6 +404,126 @@ impl<M: Middleware + 'static> UoPoolService<M> {
...
+        self.get_code_hashes(contract_addresses, code_hashes)
+            .await?;
+
+        let mempool_id = mempool_id(entry_point, &self.chain_id);
+        let user_operation_hash = user_operation.hash(entry_point, &self.chain_id);
+
+        if let Some(mempool) = self.mempools.write().get_mut(&mempool_id) {
+            match mempool.has_code_hashes(&user_operation_hash) {
+                Ok(true) => {
+                    // 2nd simulation
```
zsluedem, line 497: This is an interesting way to note it as 2nd simulation.

I think it makes more sense to have an explicit compare.

src/uopool/services/simulation.rs
```diff
@@ -443,6 +564,11 @@ impl<M: Middleware + 'static> UoPoolService<M> {
         // verify call stack
         self.call_stack(entry_point, &stake_info_by_entity, &js_trace)?;
 
+        // verify code hashes
+        // should be last verification because code hashes are stored in the 1st simulation
+        self.code_hashes(user_operation, entry_point, &js_trace)
```
zsluedem, line 569: I thought this would happen after the simulation in the bundle steps.

src/uopool/services/simulation.rs
```diff
@@ -403,6 +404,126 @@ impl<M: Middleware + 'static> UoPoolService<M> {
...
+                    });
+                }
+            }
+        }
+
+        Ok(())
+    }
+
+    fn compare_code_hashes(
+        &self,
```
zsluedem, line 443: `self` is acutally not needed here. We could make the compare into a more general method.

zsluedem (commented):
Left some comments

src/uopool/services/simulation.rs
Vid201: Which part - the whole verification of code hashes or storing code hashes after 1st simulation?

i think it make sense to include this here since it's part of the simulation.

src/uopool/services/simulation.rs
Vid201: I was also thinking of wrapping user operation, but that would mean passing user operation around with additional data that is not needed most of the time (except here).

Vid201: One option is to add flag parameter to simulate function.

1st simulation is always when adding user operation to mempool.
2nd simulation is when bundling.

We can pass true/false to the function on that places.

src/uopool/services/simulation.rs
zsluedem: Nvm. I think it is OK here.

src/uopool/services/simulation.rs
zsluedem: I don't have other ideas now. Let's solve it later when we got another good plan.

zsluedem (approved):
LGTM

PR #89, opened 2023-03-27
chore: remove unused res folder

zsluedem:
@Vid201 Does it make sense to remove this?


Vid201:
> @Vid201 Does it make sense to remove this?

Yeah, I think it does, as we are currently not supporting different chains. We can  reiterate on this after we plan to support for multiple chains/rollups (there will be probably some differences between them).

Vid201 (approved):

PR #90, opened 2023-03-27
Add change bundler mode api

zsluedem:
https://github.com/Vid201/aa-bundler/issues/86


src/bundler/server.rs
```diff
@@ -0,0 +1,35 @@
...
+
+use crate::uopool::server::{
+    bundler::{bundler_server::Bundler as BundlerServer, SetModeRequest, SetModeResponse},
+    types::{GetChainIdResponse, GetSupportedEntryPointsResponse},
+};
+
+use super::Bundler;
+
+#[async_trait]
+impl BundlerServer for Bundler {
```
Vid201, line 12: Do you think it make sense to add BundlerService (similar to how it is done for user operation mempool)? Or will this just add not needed complexity?

src/proto/bundler/bundler.proto
```diff
@@ -5,15 +5,29 @@ import "types/types.proto";
 
 package bundler;
 
-message ChainIdResponse {
-    uint64 chain_id = 1;
+enum Mode {
+    AUTO = 0;
+    MANUAL = 1;
```
Vid201, line 10: If it is auto, it would make sense to pass also bundle interval value (e..g, 10 secs or 5 secs ...).

src/bundler/server.rs
zsluedem: I am not sure I understand your questions here.

Do you mean we should remove this https://github.com/Vid201/aa-bundler/blob/main/src/proto/bundler/bundler.proto ?

Vid201: No. This is now solved, my idea was to add `BundlerManager`, like you already did (we can rename it to BundlerService to it will have same name as `UoPoolService` - https://github.com/Vid201/aa-bundler/blob/main/src/uopool/services/uopool.rs#L86).

src/bundler/mod.rs
```diff
@@ -118,17 +133,117 @@ impl<'a> Bundler<'a> {
...
+        info!("Sending bundle now");
+        let mut tx_hashes: Vec<H256> = vec![];
+        for bundler in self.bundlers.iter() {
+            info!("Sending bundle for entry point: {:?}", bundler.entry_point);
+            let tx_hash = bundler.send_next_bundle().await?;
+            tx_hashes.push(tx_hash)
+        }
+
+        // FIXME: Because currently the bundler support multiple bundler and
+        // we don't have a way to know which bundler is the one that is
```
zsluedem, line 196: @Vid201 Do you have any good ideas on addressing the multiple entry point here?

Vid201: Hmmm, I asked in Telegram about that.

bin/bundler.rs
```diff
@@ -107,16 +109,24 @@ fn main() -> Result<()> {
...
-                    let _bundler = Bundler::new(
-                        &wallet,
-                        opt.bundler_opts.beneficiary,
-                        uopool_grpc_client.clone(),
-                        opt.bundler_opts.bundle_interval,
-                        *entry_point,
-                        opt.eth_client_address.clone(),
-                    );
-                }
+                let bundler_manager = BundlerService::new(
```
Vid201, line 112: let bundler_service

bin/bundler.rs
```diff
@@ -107,16 +109,24 @@ fn main() -> Result<()> {
...
+                    wallet,
+                    opt.bundler_opts.beneficiary,
+                    uopool_grpc_client.clone(),
+                    opt.entry_points,
+                    opt.eth_client_address.clone(),
+                );
+                info!("Starting bundler manager");
```
Vid201, line 119: bundler service

src/bundler/mod.rs
Vid201: Send bundle on  each entry point.

zsluedem: It is sending on each entry point. It is just the standard EIP output  only have one transaction hash result. So I would just return one result.

zsluedem: https://github.com/eth-infinitism/account-abstraction/blob/develop/eip/EIPS/eip-4337.md#-debug_bundler_sendbundlenow

Vid201: Ohh, I get it now. The current tests are meant for only one entry point.

Let's make the function compatible with tests now (e.g., return the first element of vec of bundlers), and leave a comment in the code to fix later.

Vid201 (approved):

PR #96, opened 2023-03-30
Chore: some rename

zsluedem:
@Vid201 Sorry. I thought I pushed the changes but it turns out I didn't push the latest changes in #90 and I merged it .
These should be good with some chore change.


Vid201 (approved):
Looks good

PR #98, opened 2023-03-30
chore: remove user operations from mempool

Vid201:
This PR does the following:

- implements removing user operations from memool (gRPC)
- skip waiting for transaction receipt - we can add that later, tests are running very long if we wait for receipt (and it's not needed for tests)
- fix some gas issues when sending bundle as tx (tx reverted due to running out of gas)

### DIFF
--- a/src/uopool/services/uopool.rs
+++ b/src/uopool/services/uopool.rs
@@ -15,7 +15,7 @@ use crate::{
                 ClearResult, EstimateUserOperationGasRequest, EstimateUserOperationGasResponse,
                 EstimateUserOperationGasResult, GetAllReputationRequest, GetAllReputationResponse,
                 GetAllReputationResult, GetAllRequest, GetAllResponse, GetAllResult,
-                GetSortedRequest, GetSortedResponse, RemoveRequest, RemoveResponse,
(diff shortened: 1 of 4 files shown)


src/uopool/services/uopool.rs
```diff
@@ -160,9 +160,45 @@ where
...
+
+        if let RemoveRequest {
+            hashes,
+            ep: Some(entry_point),
+        } = req
+        {
+            let entry_point = entry_point
+                .try_into()
+                .map_err(|_| tonic::Status::invalid_argument("invalid entry point"))?;
+            let mempool_id = mempool_id(&entry_point, &self.chain_id);
```
zsluedem, line 175: I just want to say that these mempool id routing codes are duplicates in every method.  That's what I mean we need route service in https://github.com/Vid201/aa-bundler/issues/78#issuecomment-1483975167 so that we could remove dup codes here.

zsluedem: We could do that later. Just want to bring up here.

zsluedem (approved):
LGTM

src/uopool/services/uopool.rs
Vid201: Yeah, I agree. The more features we add, the more I see the problem.

We have to definitely make that change later.

Issue #70, opened 2023-03-08
Remove user operations in the pool when clients create a block with the user operartions


zsluedem:
This is a leftover in https://github.com/Vid201/aa-bundler/pull/68 .

The user operations should be removed when they are included in the blocks.

The request is in which component the bundler should keep an eye on it **the bundler** or the **memory pool**

Vid201:
This relates to #63 . I personally prefer **memory pool**, what do you think?

zsluedem:
> This relates to #63 . I personally prefer **memory pool**, what do you think?

hm.. I think that's ok. Because of the architecture of the current bundler, we need to do a lot more work to address internal component communications than other bundlers.

PR #101, opened 2023-04-06
feat: remove user operation when bundles are included in a block

zsluedem:
#70


src/uopool/services/uopool.rs
```diff
@@ -471,6 +490,71 @@ where
...
+    async fn handle_past_events(
+        &self,
+        request: tonic::Request<HandlePastEventRequest>,
+    ) -> Result<Response<HandlePastEventResponse>, tonic::Status> {
+        let req = request.into_inner();
+        let HandlePastEventRequest {
+            entry_point: entry_point_opt,
+            chain_id,
+        } = req;
+        let lastest_block = self
```
Vid201, line 502: `latest_block`

src/proto/uopool/uopool.proto
```diff
@@ -108,13 +108,21 @@ message GetSortedResponse{
...
+message HandlePastEventResponse{}
+
 service UoPool {
     rpc Add(AddRequest) returns (AddResponse);
     rpc Remove(RemoveRequest) returns (RemoveResponse);
     rpc GetChainId(google.protobuf.Empty) returns (types.GetChainIdResponse);
     rpc GetSupportedEntryPoints(google.protobuf.Empty) returns (types.GetSupportedEntryPointsResponse);
     rpc EstimateUserOperationGas(EstimateUserOperationGasRequest) returns (EstimateUserOperationGasResponse);
     rpc GetSortedUserOperations(GetSortedRequest) returns (GetSortedResponse);
+    rpc HandlePastEvents(HandlePastEventRequest) returns (HandlePastEventResponse);
```
Vid201, line 125: Response type can be replaced with `google.protobuf.Empty`

src/proto/uopool/uopool.proto
```diff
@@ -108,13 +108,21 @@ message GetSortedResponse{
     repeated types.UserOperation user_operations = 1;
 }
 
+message HandlePastEventRequest{
+    types.H160 entry_point = 1;
+    uint64 chain_id = 2;
```
Vid201, line 113: Do we need to pass chain_id? Uopool instance has chain id in struct

src/uopool/services/uopool.rs
```diff
@@ -471,6 +490,71 @@ where
...
+            .clone()
+            .get_block_number()
+            .await
+            .map_err(|e| {
+                tonic::Status::internal(format!("Getting the latest block number error: {e:?}"))
+            })?;
+        let last_block = std::cmp::max(
+            1u64,
+            lastest_block
+                .checked_sub(U64::from(1000))
```
Vid201, line 513: maybe make this value const and move to mod - so we change more easily change it later if needed.

Vid201: 1000

src/bundler/mod.rs
```diff
@@ -130,15 +131,29 @@ impl Bundler {
...
+        trace!("Prepare the transaction {tx:?} send to execution client!");
         let tx = client.send_transaction(tx, None).await?;
         let tx_hash = tx.tx_hash();
-        debug!("Send bundle with transaction: {tx:?}");
+        trace!("Send bundle with transaction: {tx:?}");
 
-        // TODO: check if bundle was included on-chain
-        // let res = tx.await?;
-        // debug!("Send bundle with receipt: {res:?}");
+        info!("Send handlePastEvents request");
```
Vid201, line 144: I am not sure if we need to wait for the event here.

The bundler should automatically listen for past events when started - and do it all the time.

Vid201 (changes requested):
Good job!

Left some comments.

src/bundler/mod.rs
zsluedem: It doesn't wait here. It would just simply check the history events. It is possible that the bundles just sent are not included in blocks at this event handling.

Vid201: Ok, my bad, then it makes sense.

But we should have `handle_past_events` also for bundles submitted by other bundlers (I guess we can call handle past events after each new block is added - and handle reorgs also in the future).

Vid201 (approved):

src/bundler/mod.rs
Vid201: We can create issue to handle these two things later.

Issue #100, opened 2023-04-03
Replace UserOperation with higher gas


zsluedem:
Currently we are missing the part which the bundler could replace the UserOperation if the sender provide higher gas fee and reject if the sender provide lower or not high enough for threshold

https://github.com/eth-infinitism/bundler-spec-tests/blob/master/tests/bundle/test_bundle.py#L44-L52

PR #108, opened 2023-04-12
Replace user operation

Vid201:
Closese #100

### DIFF
--- a/src/bundler/mod.rs
+++ b/src/bundler/mod.rs
@@ -1,6 +1,6 @@
 pub mod service;
 
-use std::{net::SocketAddr, sync::Arc, time::Duration};
+use std::{net::SocketAddr, sync::Arc, thread, time::Duration};
 
 use clap::Parser;
 use ethers::{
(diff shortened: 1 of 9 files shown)


src/bundler/mod.rs
```diff
@@ -138,6 +138,9 @@ impl Bundler {
         let tx_hash = tx.tx_hash();
         trace!("Send bundle with transaction: {tx:?}");
 
+        // wait some time, otherwise sometimes events for this bundle are not yet detected
+        thread::sleep(Duration::from_millis(25));
```
zsluedem, line 142: use `tx.await` ?

src/uopool/sanity_check.rs
```diff
@@ -13,6 +13,14 @@ use ethers::{
 
 use super::UoPool;
 
+const MAX_UOS_PER_UNSTAKED_SENDER: usize = 4;
+const GAS_INCREASE_PERC: u64 = 10;
```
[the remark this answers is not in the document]
Vid201: Yeah, they are not as far as I know (only in bundler spec tests).

This parameters could probably be configurable by each bundler in the future though, I don't think they have to be explicitly defined in the ERC-4337.

I think we can have it set this way for now, and see later what to do with them (or just leave them like that).

src/bundler/mod.rs
Vid201: Yeah, that's better. I updates the polling interval for the pending transaction from 7 secs to 0.1, so 1 test (that contains sending bundle) is not running for 7 seconds +.

zsluedem (approved):
LGTM

PR #110, opened 2023-04-25
Refactor project to workspace

Vid201:
- Refactor project to workspace
- Add licenses to the project


Cargo.toml
```diff
@@ -1,73 +1,20 @@
...
+    "crates/contracts",
+    "crates/grpc",
+    "crates/primitives",
+    "crates/rpc",
+    "crates/uopool",
+    "tests",
+]
+default-members = ["bin/bundler"]
+
+[profile.debug-fast]
```
zsluedem, line 14: I am curious what this means. I could not google any good results on it. Maybe you could provide some links or description?

zsluedem, line 14: I am curious what this means. I could not google any good results on it. Maybe you could provide some links or description?

zsluedem (approved):
This is great work!! I only have one question. It should be good to merge.

Cargo.toml
Vid201: This enables us to to do: `cargo build --profile debug-fast` and have debug build for debugging with high optimizations (as in release).

Link (example): https://doc.rust-lang.org/cargo/reference/profiles.html#custom-profiles

I saw this in reth and it looked useful: https://github.com/paradigmxyz/reth/blob/main/Cargo.toml

PR #116, opened 2023-05-01
Run bundler test spec

zsluedem:


.github/workflows/bundle_spec_test.yml
```diff
@@ -0,0 +1,43 @@
+on:
+  workflow_run:
+    workflows: ["Create and publish a Docker image"]
+    types: [completed]
+
+name: run bundler test spec
```
Vid201, line 6: Change to "Run bundler spec tests"

.github/workflows/bundle_spec_test.yml
```diff
@@ -0,0 +1,43 @@
...
+      - run: cd bundler-spec-tests; git checkout
+        name: "re-checkout bundler-spec-tests (on top of cache)"
+
+        #for faster "update-deps" for spec
+      - run: yarn --cwd bundler-spec-tests/spec remove gatsby || echo "already removed"
+
+      - run: "cd bundler-spec-tests && pdm install && pdm update-deps"
+        name: update submodules of bundler-spec-tests
+
+      - run: ./runall.sh launchers/aabundler-launcher.sh
```
Vid201, line 42: Shouldn't this be `./runall.sh launchers/aa-bundler-rust-launcher.sh` ?

Vid201 (changes requested):
Looks good, makes sense to build docker image only on new tags!

.github/workflows/bundle_spec_test.yml
zsluedem: You are right

Vid201 (approved):

PR #118, opened 2023-05-03
Test ci 

zsluedem:


Vid201:
When I created PR for updating entry point, bundler spec tests workflow was started. Is this normal? Shouldn't it start only when we create new Docker image?

Also, let's rename file from `bundle_spec_test.yml` to `bundler_spec_tests.yml`.

@zsluedem

zsluedem:
@Vid201 Ok. I think we need to make sure of several things first. 

1. **Do we need to run the bundler-spec-tests when a pr is created before merging?**
2. If 1  is yes, we should run it with the docker image(with bundler-test-executor) or run the tests in spec repo(https://github.com/eth-infinitism/bundler-spec-tests) directly.

Currently the `github.actor` would be the pr author and the GITHUB_TOKEN would be the token from the pr author which means I could not create the image to your ghcr docker hub unlesss you [grant the permission to me](https://docs.github.com/en/packages/learn-github-packages/about-permissions-for-github-packages#granular-permissions-for-userorganization-scoped-packages) (same to other contributor).

Vid201:
> @Vid201 Ok. I think we need to make sure of several things first.
> 
> 1. **Do we need to run the bundler-spec-tests when a pr is created before merging?**
> 2. If 1  is yes, we should run it with the docker image(with bundler-test-executor) or run the tests in spec repo(https://github.com/eth-infinitism/bundler-spec-tests) directly.
> 
> Currently the `github.actor` would be the pr author and the GITHUB_TOKEN would be the token from the pr author which means I could not create the image to your ghcr docker hub unlesss you [grant the permission to me](https://docs.github.com/en/packages/learn-github-packages/about-permissions-for-github-packages#granular-permissions-for-userorganization-scoped-packages) (same to other contributor).

We should run **bundler-spec-tests** when a PR is created before merging.

I think we should run it directly with repo **bundler-spec-tests**. In case the tests are passed, we can create a new docker image and publish it (dev tag and latest) when merging.

I think you have correct permissions, the tag was messed up IMO.

Screenshot from 2023-05-11 11-14-44

PR #132, opened 2023-05-23
fix: fix auto bundling

zsluedem:


Vid201 (approved):
LGTM!

One observation: would it make sense to move auto and manual bundling logic to crate bundler and pass it in the constructor (so when you start the bundler you can specify the bundling mode)?

@zsluedem

zsluedem:
@Vid201  Current design which allows changing the mode in the runtime is because of the [debug_bundler_setBundlingMode](https://github.com/eth-infinitism/account-abstraction/blob/develop/eip/EIPS/eip-4337.md#-debug_bundler_setbundlingmode) required for the spec tests. I think it makes sense to specify it when we start the bundler.

PR #135, opened 2023-05-25
Various fixes

Vid201:
This PR fixes bugs that were found during testing with trampoline and candide wallet on Sepolia.

I also changed CI for publishing image so image is build only when something is pushed to main branch.

### DIFF
--- a/crates/uopool/src/utils.rs
+++ b/crates/uopool/src/utils.rs
@@ -67,7 +67,7 @@ impl Overhead {
                 })
                 .sum::<u128>(),
         );
-        let length_in_word = ((user_operation_packed.len() + 31) as f64) / (32 as f64);
+        let length_in_word = ((user_operation_packed.len() + 31) as f64) / 32_f64;
         u256_from_f64_saturating(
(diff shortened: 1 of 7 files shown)


zsluedem (approved):
LGTM

crates/uopool/src/utils.rs
```diff
@@ -67,12 +67,11 @@ impl Overhead {
...
         );
+        let length_in_word = ((user_operation_packed.len() + 31) as f64) / 32_f64;
         u256_from_f64_saturating(
             (self.fixed.as_u128() as f64) / (self.bundle_size.as_u128() as f64)
-                + (call_data_cost
-                    + self.per_user_op
-                    + self.per_user_op_word * user_operation_packed.len())
-                .as_u128() as f64,
+                + ((call_data_cost + self.per_user_op).as_u128() as f64)
+                + (self.per_user_op_word.as_u128() as f64) * length_in_word,
```
zsluedem, line 74: we need to review these kinds of converting someday. These can be attack vectors and unsafe sometimes.

Vid201: Yeah, agree.

PR #136, opened 2023-05-26
Add web3_clientversion rpc

zsluedem:


crates/rpc/src/eth.rs
```diff
@@ -22,6 +22,8 @@ use tracing::{debug, trace};
 
 use crate::eth_api::EthApiServer;
 
+const SUPPORT_ENTRYPOINT_VERSION: &str = "0.6.0";
```
Vid201, line 25: I would move this const value to primitives crate (lib.rs).

crates/rpc/src/eth_api.rs
```diff
@@ -38,4 +38,7 @@ pub trait EthApi {
         &self,
         user_operation_hash: String,
     ) -> RpcResult<Option<UserOperationByHash>>;
+
+    #[method(name = "web3_clientVersion")]
```
Vid201, line 42: I didn't try it out, but since this method is in eth namespace, shouldn't this result in eth_web3_clientVersion, instead of only web3_cclientVersion?

Vid201 (changes requested):

crates/rpc/src/eth_api.rs
zsluedem: OH. You are right.

crates/rpc/src/custom_api.rs
```diff
@@ -0,0 +1,7 @@
+use jsonrpsee::{core::RpcResult, proc_macros::rpc};
+
+#[rpc(server)]
+pub trait CustomApi {
```
Vid201, line 4: I think it would be better to have web3.rs and web3_api.rs instead of custom, to follow the same pattern.

Vid201 (approved):

PR #137, opened 2023-05-27
feat: uopool unsafe

Vid201:


crates/primitives/src/uopool.rs
```diff
@@ -0,0 +1,5 @@
+#[derive(clap::ValueEnum, Clone, Copy, Debug, PartialEq, Eq)]
+pub enum Mode {
+    Auto,
```
zsluedem, line 3: `Auto` seems ambiguous to me. Any better words?

zsluedem (approved):
Generally good

crates/primitives/src/uopool.rs
Vid201: I was thinking about "Standard". Do you have any other suggestion?

Vid201:
> Generally good

We will need to refactor verification of user operations for alt mempools, since we will have then normal, unsafe, altmempool1, altmempool2 ...

PR #138, opened 2023-06-02
Fix call gas limit

Vid201:
This PR fixes estimation for call gas limit.

Code is a little bit messy, gonna refactor with some other things in the subsequent PR.


crates/uopool/src/canonical/sanity_check.rs
```diff
@@ -459,6 +493,7 @@ mod tests {
 
     use super::*;
 
+    #[ignore]
```
zsluedem, line 496: https://github.com/Vid201/aa-bundler/actions/runs/5166205480/jobs/9306373395
This is stopping me merge this pr.

zsluedem (approved):
Looks good to me

crates/uopool/src/canonical/sanity_check.rs
zsluedem: It would be great to mock the case locally instead of relying on external source in test cases.

Vid201: Yeah, agree. I was thinking of removing this test case for now and revisiting it later. You can remove it in your PR.

PR #148, opened 2023-06-12
feat: route requests to execution client

Vid201:
Since it's not possible to handle unknown methods directly with jsonrpsee (https://github.com/paritytech/jsonrpsee/issues/984), we need custom middleware.


crates/rpc/src/middleware.rs
```diff
@@ -0,0 +1,103 @@
+use hyper::{Body, Request, Response};
```
zsluedem, line 1: very good solution!

zsluedem (approved):
LGTM

PR #150, opened 2023-06-16
Update stake/reputation

Vid201:
Error messages were adapted based on the test stake rules from bundler spec tests.


zsluedem (approved):
LGTM

PR #156, opened 2023-06-28
ssz support for user operation

zsluedem:


crates/primitives/src/user_operation.rs
```diff
@@ -405,6 +410,274 @@ pub struct UserOperationGasEstimation {
...
+impl ssz_rs::Sized for UserOperation {
+    fn is_variable_size() -> bool {
+        true
+    }
+    fn size_hint() -> usize {
+        0
+    }
+}
+
+impl ssz_rs::Serialize for UserOperation {
```
zsluedem, line 452: @Vid201 I am handwriting this because I don't want to introduce another type of conversion. Do you have better ideas? Maybe we should wrap `Bytes`, `U256` these kinds of types in our crates?

Vid201: Hmm, I can't think of better solution at the moment. I think this is fine for now.

Should we add some 1 or 2 tests for ssz serialize/deserialize?

Vid201 (approved):

crates/primitives/src/user_operation.rs
zsluedem: Yes. We should

crates/primitives/src/user_operation.rs
```diff
@@ -484,4 +758,44 @@ mod tests {
                 .into()
         );
     }
+
+    #[test]
+    fn user_operation_ssz() {
```
zsluedem, line 763: @Vid201 add a test case

Vid201: Looks good

Vid201 (approved):

PR #162, opened 2023-07-15
Abstract sanity check and simulation rules

Vid201:
This PR abstracts sanity check and simulation  results, making it easier to implement alternative mempools when they arrive. It's easy to add new rules if necessary and decide which rules you want to use.

I added 3 steps to user operation validator:
- sanity -> basic sanity checks
- simulation -> calls simulateValidation but without tracing
- simulation trace -> call simulateValidation with JS tracing

Feel free to suggest any changes (outcome and helpers of validation could probably be improved in the future, but I didn't think of a better solution at the moment).


crates/uopool/src/validate/mod.rs
```diff
@@ -0,0 +1,110 @@
...
+    stake_info: Option<[StakeInfo; NUMBER_LEVELS]>,
+    code_hashes: Option<Vec<CodeHash>>,
+}
+
+#[async_trait::async_trait]
+pub trait SimulationTraceCheck<M: Middleware>: Send + Sync {
+    async fn check_user_operation(
+        &self,
+        uo: &UserOperation,
+        helper: &mut SimulationTraceHelper<M>,
```
zsluedem, line 108: The helper here is mutable. Does it matter if the `SimulationTraceCheck` is checked in a different order?

What I mean is  the check `A-> B -> C`

A and C doesn't mutate helper while B mutate the helper. If I reorder the check like `B -> A -> C`. Does that matter to A while some state is changed in B?

zsluedem: Same question to other checks.

crates/grpc/src/uopool.rs
```diff
@@ -60,7 +78,7 @@ where
 
         let res = {
```
zsluedem, line 79: The res here should be `Result<T,E>` type. Should I consider the error situation here and return Error status when validation failed?

zsluedem (commented):
Generally looks good.Left some comments.

crates/grpc/src/uopool.rs
Vid201: Not sure what you mean here. The error is handled a little bit later (match): https://github.com/Vid201/aa-bundler/pull/162/files#diff-24e5e1943959520a80d3685779d0161bf51ac80c4855df204fefc53db2c6497bR92

crates/uopool/src/validate/mod.rs
Vid201: The order doesn't matter - can be in any way.

It is mutable because for different reason: for example it contains `code_hashes`, and multiple checks can do something with `code_hashes`. So if one check fetches `code_hashes` from eth_client, the other checks doesn't have to repeat the same calls (saving some time).

crates/grpc/src/uopool.rs
zsluedem: I see. Why not handled immediately?

crates/uopool/src/uopool.rs
```diff
@@ -101,17 +93,34 @@ impl<M: Middleware + 'static> UoPool<M> {
...
+            )
+            .await
+    }
+
+    /// Adds a single validated user operation into the pool
     pub async fn add_user_operation(
         &mut self,
         uo: UserOperation,
-        res: Option<Result<VerificationResult, VerificationError>>,
+        res: Option<Result<UserOperationValidationOutcome, ValidationError>>,
```
zsluedem, line 116: Related to https://github.com/Vid201/silius/pull/162/files#r1275917583.
Why not take `Option<UserOperationValidationOutcome>`

crates/uopool/src/uopool.rs
```diff
@@ -101,17 +93,34 @@ impl<M: Middleware + 'static> UoPool<M> {
...
     ) -> Result<UserOperationHash, AddError> {
         let res = match res {
             Some(res) => res?,
```
zsluedem, line 119: I don't get why the error handling postpond to here.

crates/grpc/src/uopool.rs
```diff
@@ -332,17 +353,45 @@ pub async fn uopool_service_run(
...
+            let entry_point = EntryPoint::<Provider<Http>>::new(eth_client.clone(), ep);
+
+            let mut validator =
+                StandardUserOperationValidator::new(eth_client.clone(), entry_point.clone(), chain)
+                    .with_sanity_check(SenderOrInitCode {})
+                    .with_sanity_check(VerificationGas {
+                        max_verification_gas,
+                    })
+                    .with_sanity_check(Paymaster {})
+                    .with_sanity_check(CallGas {})
```
zsluedem, line 365: A nitpick. With the definition `pub struct CallGas;`, here you could just use `CallGas` instead of `CallGas {}`.

crates/uopool/src/mempool.rs
```diff
@@ -30,6 +30,7 @@ pub trait Mempool: Debug {
     fn get(&self, uo_hash: &UserOperationHash) -> Result<Option<UserOperation>, Self::Error>;
     fn get_all_by_sender(&self, addr: &Address) -> Self::UserOperations;
     fn get_number_by_sender(&self, addr: &Address) -> usize;
+    fn get_prev_by_sender(&self, uo: &UserOperation) -> Option<UserOperation>;
```
zsluedem, line 33: This method looks identical for two implementations. We could define a `MempoolExtension` trait and define this as a extension method. In that case, you don't have to replicate identical implementations.

zsluedem: https://rust-lang.github.io/rfcs/0445-extension-trait-conventions.html here is some links for reference.

The idea is to add methods to an existing type. 
We could define
```rust
trait MempoolExt{
    fn get_prev_by_sender(&self, uo: &UserOperation) -> Option<UserOperation>;
}

impl<T: Mempool> MempoolExt for T {
    fn get_prev_by_sender(&self, uo: &UserOperation) -> Option<UserOperation> {
        self.get_all_by_sender(&uo.sender)
            .into_iter()
            .filter(|prev_uo| prev_uo.nonce == uo.nonce - 1)
            .max_by_key(|prev_uo| prev_uo.max_priority_fee_per_gas)
    }
}
```

Then you could directly use `get_prev_by_sender` for all the implementation of Mempool when you import the package with `use uopool::MempoolExt`. In this case, you don't have to replicate the same implementation twice but only one time.

zsluedem: Actually there is another easier way to do that without defining another trait. Because the `Mempool` trait is actually defined within the our crates, We could just 
```rust
pub trait Mempool: Debug {
    type UserOperations: IntoIterator<Item = UserOperation>;
    type CodeHashes: IntoIterator<Item = CodeHash>;
    type Error;
    fn add(
        &mut self,
        uo: UserOperation,
        ep: &Address,
        chain_id: &U256,
    ) -> Result<UserOperationHash, Self::Error>;
    fn get(&self, uo_hash: &UserOperationHash) -> Result<Option<UserOperation>, Self::Error>;
    fn get_all_by_sender(&self, addr: &Address) -> Self::UserOperations;
    fn get_number_by_sender(&self, addr: &Address) -> usize;
    fn get_prev_by_sender(&self, uo: &UserOperation) -> Option<UserOperation>{
        self.get_all_by_sender(&uo.sender)
            .into_iter()
            .filter(|prev_uo| prev_uo.nonce == uo.nonce - 1)
            .max_by_key(|prev_uo| prev_uo.max_priority_fee_per_gas)
    }
    fn has_code_hashes(&self, uo_hash: &UserOperationHash) -> Result<bool, Self::Error>;
    fn set_code_hashes(
        &mut self,
        uo_hash: &UserOperationHash,
        hashes: &Self::CodeHashes,
    ) -> Result<(), Self::Error>;
    fn get_code_hashes(&self, uo_hash: &UserOperationHash) -> Self::CodeHashes;
    fn remove(&mut self, uo_hash: &UserOperationHash) -> Result<(), Self::Error>;
    // Get UserOperations sorted by max_priority_fee_per_gas without dup sender
    fn get_sorted(&self) -> Result<Self::UserOperations, Self::Error>;
    fn get_all(&self) -> Self::UserOperations;
    fn clear(&mut self);
}
```

zsluedem: These are 2 ways to keep DRY. Extension method make extend external trait possible but in our case the trait is not external so the second way could work too.

crates/grpc/src/uopool.rs
Vid201: For `uo_pool.validate_user_operation` we first get immutable reference `uo_pool` and we can have multiple of them. When adding user operation, we need to get mutable which can be only one. That's the reason to split validating and adding. Does this make sense?

crates/uopool/src/mempool.rs
Vid201: Ok, got it. I decided to go for second option.

crates/grpc/src/uopool.rs
Vid201: I get your concern now. The thing I wrote is not related to the problem, I'll make a change.

zsluedem (approved):
Thanks for addressing the comments. Very nice work! :+1:

crates/uopool/src/validate/mod.rs
zsluedem: Ok. Got you.

PR #165, opened 2023-07-18
[WIP]Send bundles (txs) to Flashbots

da-bao-jian:
## Description

Sending bundles to Flashbot's Relay has the benefit of avoiding front-run and sandwich attacks. Currently, the `bundler` only supports sending transaction bundles through the public mempool, which is vulnerable to the said attacks above. As the volume of `UserOperations` picks up, each bundle submitted will likely generate more MEV, hence attracting more MEV bots. 

Allowing the bundlers to submit bundles through Flashbots has the following benefits:
* Avoiding the malicious MEV attacks from the public mempool
* Paving the way to support more MEV-related features, such as MEV-protect and MEV-Share 

## Things changed 
[crates/bundler/.env.example](https://github.com/Vid201/aa-bundler/compare/main...qi-protocol:aa-bundler:main#diff-6815648c1f8820dff3f99967067dd164e85c47303df3b0c6e29b897afc22e2a4)
* Added `FLASHBOTS_IDENTIFIER` as it is required by Flashbots(see [How to send your first Flashbots bundle](https://docs.flashbots.net/flashbots-auction/searchers/quick-start#how-to-send-your-first-flashbots-bundle))

[crates/bundler/Cargo.toml](https://github.com/Vid201/aa-bundler/compare/main...qi-protocol:aa-bundler:main#diff-e1496d17ceff40a375ac4c88fbdf7ec9d2e2e4161c6c95bd9c8d3bc24fc5b5a9)
* Added `ethers-flashbot`, `url`, `tokio` crates to support functionalities in the `send_next_bundle_flashbots` function

 [crates/bundler/src/bundler.rs](https://github.com/Vid201/aa-bundler/compare/main...qi-protocol:aa-bundler:main#diff-1814e47795f509169ce64386311f879aaa213ab63ac9b393da08c88cfec93613)
* Added the `RELAY_ENDPOINTS` for most of the commonly used block builder endpoint
* Added the `send_next_bundle_flashbots` function
  

## Pending Issues
There is no good way to test sending bundle using Flashbots unless a remote endpoint is used, such as Flashbots' Bundle Relay API on Goerli or Sepolia

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,10 +14,10 @@ run-create-wallet:
 	cargo run --release --bin create-wallet -- --output-path ${HOME}/.silius
 
 run-silius-debug:
-	cargo run --release -- --eth-client-address http://127.0.0.1:8545 --mnemonic-file ${HOME}/.silius/0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266 --beneficiary 0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266 --gas-factor 600 --min-balance 1 --entry-points 0x5FF137D4b0FDCD49DcA30c7CF57E578a026d2789 --min-stake 1 --min-unstake-delay 0 --min-priority-fee-per-gas 0 --max-verification-gas 1500000 --rpc-api eth,debug
+	cargo run --release -- --eth-client-address http://127.0.0.1:8545 --mnemonic-file ${HOME}/.silius/0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266 --beneficiary 0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266 --gas-factor 600 --min-balance 1 --entry-points 0x5FF137D4b0FDCD49DcA30c7CF57E578a026d2789 --min-stake 1 --min-unstake-delay 0 --min-priority-fee-per-gas 0 --max-verification-gas 1500000 --rpc-api eth,debug,web3
 
 run-silius-debug-mode:
-	RUST_LOG=silius=TRACE cargo run --profile debug-fast -- --eth-client-address http://127.0.0.1:8545 --mnemonic-file /home/vid/.silius/0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266 --beneficiary 0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266 --gas-factor 600 --min-balance 1 --entry-points 0x5FF137D4b0FDCD49DcA30c7CF57E578a026d2789 --min-stake 1 --min-unstake-delay 0 --min-priority-fee-per-gas 0 --max-verification-gas 1500000 --rpc-api eth,debug
+	RUST_LOG=silius=TRACE cargo run --profile debug-fast -- --eth-client-address http://127.0.0.1:8545 --mnemonic-file /home/vid/.silius/0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266 --beneficiary 0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266 --gas-factor 600 --min-balance 1 --entry-points 0x5FF137D4b0FDCD49DcA30c7CF57E578a026d2789 --min-stake 1 --min-unstake-delay 0 --min-priority-fee-per-gas 0 --max-verification-gas 1500000 --rpc-api eth,debug,web3
 
 fetch-thirdparty:
 	git submodule update --init
--- a/bin/silius/src/cli.rs
+++ b/bin/silius/src/cli.rs
@@ -48,7 +48,7 @@ pub struct RpcServiceOpts {
     #[clap(long, default_value = "127.0.0.1:3000")]
     pub rpc_listen_address: String,
 
-    #[clap(long, value_delimiter=',', default_value = "eth", value_parser = ["eth", "debug"])]
+    #[clap(long, value_delimiter=',', default_value = "eth", value_parser = ["eth", "debug", "web3"])]
     pub rpc_api: Vec<String>,
 
     #[clap(long, value_delimiter = ',', default_value = "*")]
--- a/bin/silius/src/silius-rpc.rs
+++ b/bin/silius/src/silius-rpc.rs
@@ -45,7 +45,9 @@ async fn main() -> Result<()> {
         .with_proxy(opt.eth_client_address)
         .with_cors(opt.rpc_opts.cors_domain);
 
-    server.add_method(Web3ApiServerImpl {}.into_rpc())?;
+    if api.contains("web3") {
+        server.add_method(Web3ApiServerImpl {}.into_rpc())?;
+    }
 
     let uopool_grpc_client =
         UoPoolClient::connect(format!("http://{}", opt.uopool_grpc_listen_address)).await?;
--- a/bin/silius/src/silius.rs
+++ b/bin/silius/src/silius.rs
@@ -156,7 +156,9 @@ fn main() -> Result<()> {
                             let mut server = JsonRpcServer::new(opt.rpc_opts.rpc_listen_address.clone()).with_proxy(opt.eth_client_address)
                             .with_cors(opt.rpc_opts.cors_domain);
 
-                            server.add_method(Web3ApiServerImpl{}.into_rpc())?;
+                            if api.contains("web3") {
+                                server.add_method(Web3ApiServerImpl{}.into_rpc())?;
+                            }
 
                             if api.contains("eth") {
                                 server.add_method(
--- a/bundler-spec-test/launcher.sh
+++ b/bundler-spec-test/launcher.sh
@@ -21,7 +21,7 @@ case $1 in
         --min-stake 1 \
         --min-unstake-delay 0 \
         --min-priority-fee-per-gas 0 \
-        --rpc-api eth,debug \
+        --rpc-api eth,debug,web3 \
         --max-verification-gas 1500000 & echo $! > bundler.pid
     popd
 	cd @account-abstraction && yarn deploy --network localhost


crates/bundler/src/bundler.rs
```diff
@@ -1,13 +1,36 @@
...
+    },
 };
+use ethers_flashbots::{BundleRequest, FlashbotsMiddleware, PendingBundleError::BundleNotIncluded};
 use std::{sync::Arc, time::Duration};
 use tracing::{info, trace};
+use url::Url;
+use tokio::task::JoinHandle;
+use std::env;
+
+const RELAY_ENDPOINTS: &[(&str, &str)] = &[
```
zsluedem, line 20: It would be great to make the entrypoint in config. Maybe we could reuse the exist entrypoint config and add an additional flag to enable flashbot api.

crates/bundler/src/bundler.rs
```diff
@@ -78,4 +101,142 @@ impl Bundler {
 
         Ok(tx_hash)
     }
+
+    // TODO: add more relay endpoints support
+    /// Send a bundle as Flashbots bundles
+    #[allow(clippy::needless_return)]
+    #[allow(clippy::clone_double_ref)]
+    pub async fn send_next_bundle_flashbots(&self, uos: &Vec<UserOperation>, test: Option<bool>) -> anyhow::Result<H256> {
```
zsluedem, line 109: The `test` param could be removed, here. Move this kind of config in the entrypoint config.

crates/bundler/src/bundler.rs
```diff
@@ -78,4 +101,142 @@ impl Bundler {
...
+            if let Some(revert) = &tx.revert { 
+                return Err(anyhow::anyhow!(
+                    "Transaction failed simulation with revert: {:?}",
+                    revert
+                ));
+            }
+        };
+
+        // Send the Flashbots bundle and check for status
+        let handle: JoinHandle<Result<(bool, H256), anyhow::Error>> = tokio::spawn(
```
zsluedem, line 204: What is the reason to spawn a tokio thread here?

zsluedem (changes requested):
Left some comments

crates/bundler/src/bundler.rs
da-bao-jian: The requirements I had was this
-the bundler should be able to send the bundle to relays while processing new UserOperations from the alt mempool. To not block the main thread processing the latest `UserOperations`, we need a thread that handles sending and monitoring the status bundles that have been sent
-the bundler should be able to send the same bundle to multiple relays in the future to maximize the chances of inclusion. Sending and monitoring status to multiple relays is a quite complex operation, hence should not block the main thread. When we start supporting multiple relays, we can change this part of the code to something like this
```
let handles: Vec<_> = RELAY_ENDPOINTS
        .len()
        .map(|i| {
            tokio::spawn(async move {
                println!("Starting task {}", i);
                // sending bundles
                println!("Completed task {}", i);
            })
        })
        .collect();

    for handle in handles {
        let _ = handle.await;
    }
```

crates/bundler/Cargo.toml
```diff
@@ -17,3 +17,6 @@ aa-bundler-primitives = { path = "../primitives" }
 anyhow = "1"
 ethers = { workspace = true }
 tracing = { workspace = true }
+ethers-flashbots = "0.13.1"
```
Vid201, line 20: sort deps

crates/bundler/src/bundler.rs
```diff
@@ -78,4 +101,142 @@ impl Bundler {
...
+        let client = Arc::new(
+            SignerMiddleware::new(
+                fb_middleware,
+                self.wallet.signer.clone(),
+            )
+        );
+
+        let ep = EntryPointAPI::new(self.entry_point, client.clone());
+
+        let nonce = client
```
Vid201, line 147: some code is repeated in this function and normal send bundle, we could move some logic into new functions

crates/bundler/src/bundler.rs
Vid201, line 20: +1 for moving this information in some other file and adding flashbots cmd param: --bundle normal/flashbots (or something like that, enum)

Vid201 (changes requested):

crates/bundler/src/bundler.rs
```diff
@@ -78,4 +101,142 @@ impl Bundler {
...
+        };
+
+        info!("Creating a new bundle with {} user operations", uos.len());
+        trace!("Bundle content: {uos:?}");
+
+        let provider = Provider::<Http>::try_from(self.eth_provider_address.clone())?;
+
+        let _bundle_signer = env::var("FLASHBOTS_IDENTIFIER").expect("FLASHBOTS_IDENTIFIER environment variable is not set");
+
+        let bundle_signer = _bundle_signer.parse::<LocalWallet>()?;
```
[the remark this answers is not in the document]
da-bao-jian: The problem with this is that Flashbots requires a separate signer(see [this](https://docs.flashbots.net/flashbots-auction/searchers/quick-start#how-to-send-your-first-flashbots-bundle))

Should I add the Flashbots' identifier to the wallet as well?

da-bao-jian:
added the changes accordingly. 

Only thing is that:
* when run `make lint` the `bin/bundler/Cargo.toml` still shows dep is not sorted and not formatted.

crates/bundler/Cargo.toml
da-bao-jian: what are the formatting and sorting rules for Cargo.toml

Even if I sorted the deps alphabetically, it when run `make lint` it still says not sorted

Vid201: Hmm, it should be alphabetically and nothing else. Is there any error text?

da-bao-jian:
```
error: Cargo.toml for bundler is not formatted
Checking bundler...
error: Dependencies for bundler are not sorted
Checking contracts...
Checking grpc...
Checking primitives...
Checking rpc...
Checking uopool...
Checking examples...
error: Cargo.toml for examples is not formatted
```

If you check the `example/Cargo.tomal`, and `bin/bundler/Cargo.toml` in the latest commit, they should be alphabetically sorted.

examples/Cargo.toml
```diff
@@ -21,3 +21,11 @@ tokio = { workspace = true }
 [[example]]
 name = "user-operation"
 path = "user_operation.rs"
+
+
+[dependencies]
+jsonrpsee = {version = "0.18.2", features = ["server", "http-client"]}
```
Vid201, line 27: Is not formatted error is because features are not sorted and there is no whitespace between `{version` -> `{ version` and `"http-client"]}` -> `"http-client"] }`

crates/bundler/Cargo.toml
Vid201: It's not sorted because tokio should be before tracing

Vid201 (changes requested):

examples/Cargo.toml
da-bao-jian: Thanks for pointing it out! 

I've added `cargo sort --workspace` command to the `Makefile` in the latest commit. Otherwise, it's a pain in the ass to sort & format deps manually.

Vid201: Yeah, that makes sense, cool

da-bao-jian:
For the latest update, other than addressing the previous issues, I also added some comments for Rustdoc generation. Let me know what you think so that I could know if it's the correct format to go about adding rustdocs to other files. @Vid201

Vid201:
> For the latest update, other than addressing the previous issues, I also added some comments for Rustdoc generation. Let me know what you think so that I could know if it's the correct format to go about adding rustdocs to other files. @Vid201

Hey, I checked and it looks good, it would be perfect to have that

bin/bundler/src/bundler.rs
```diff
@@ -98,9 +101,22 @@ fn main() -> Result<()> {
                     }
                 }
 
-                let wallet = Wallet::from_file(opt.mnemonic_file.clone(), &chain_id)
+                let wallet: Wallet;
+                if opt.fb_mnemonic_file.is_some() {
```
Vid201, line 105: Is there any specific reason for having different mnemonic phrase for flashbots?

crates/primitives/src/consts.rs
```diff
@@ -25,3 +25,18 @@ pub mod entities {
     pub const ACCOUNT: &str = "account";
     pub const PAYMASTER: &str = "paymaster";
 }
+
+/// Builder JSON-RPC Endpoints
+pub mod relay_endpoints {
```
Vid201, line 30: Just a suggestion: rename to `flashbots_relay_endpoints`

Vid201 (changes requested):

crates/grpc/src/bundler.rs
```diff
@@ -49,6 +49,7 @@ impl BundlerService {
         Ok(uos)
     }
 
+    // TODO: add send bundle to flashbots
```
Vid201, line 52: add CLI mode: `./bundler --send-bundle flashbots` or `./bundler --send-bundle eth-client`

bin/bundler/src/bundler.rs
da-bao-jian: I was following [the guideline of Flashbots](https://docs.flashbots.net/flashbots-auction/searchers/quick-start#how-to-send-your-first-flashbots-bundle). In particular, it shouldn't store any funds

```
When you send bundles to Flashbots you will sign them with a private key 
so that we can establish identity for searchers and establish a reputation for them over time. 
This private key does not store funds and is not the primary private key you use for executing transactions. 
```

Vid201: Yeah, I get that. But we can use different derivation paths and single seed phrase (that ways bundler admin don't have to store two seed phrases):

 "m/44'/60'/0'/0/0" and  "m/44'/60'/0'/0/1". First for executing transactions and the second for flashbots identity.

`ethers-rs` allows changing derivation paths.

da-bao-jian: Oh, nice. I didn't know that was possible. 

I'll make the change

crates/grpc/src/bundler.rs
da-bao-jian: For this, here's how I could implement it

add the `send-bundle` command in `bin/bundler/bundler`. If `flashbots`, `bundler` should use the `send_next_bundle_flashbots` method by default. There are three places I could indicate the intent to send bundle via `flashbots`
* adding an additional field to the [send_bundle](https://github.com/Vid201/aa-bundler/blob/356cee4e4c7b7aa0bee02be3280e9cc928b063df/crates/grpc/src/bundler.rs#L52) function
* adding an additional field to the [Bundler](https://github.com/Vid201/aa-bundler/blob/356cee4e4c7b7aa0bee02be3280e9cc928b063df/crates/bundler/src/bundler.rs#L13) struct;
* adding an additional field to the [bundler_service_run](https://github.com/Vid201/aa-bundler/blob/356cee4e4c7b7aa0bee02be3280e9cc928b063df/crates/grpc/src/bundler.rs#L193)

Let me know which is the best option

da-bao-jian:
As for testing, do you have any suggestions? 

I could use `flashbots` builder on Goerli, but this requires Goerli ETH. In other words, not everyone is able to pass the test if he/she doesn't have Goerli ETH.

crates/grpc/src/bundler.rs
Vid201: I would make `SendBundleMode` enum and add an additional field to the Bundler struct. Enum is parsed from CLI and passed as an additiona field to the bundler_service_run which pushes it to the Bundler struct constructor.

Vid201:
> As for testing, do you have any suggestions?
> 
> I could use `flashbots` builder on Goerli, but this requires Goerli ETH. In other words, not everyone is able to pass the test if he/she doesn't have Goerli ETH.
> 
> Alternatively, I could send an `eth_simulate` request. If the payload indicates success, it'll pass the test. I believe simulation requires no real payment.

Makes sense. I think simulating is enough for the tests. Do you mean this endpoint: https://docs.flashbots.net/flashbots-auction/searchers/advanced/rpc-endpoint#eth_callbundle ?

We can also have a test that is skipped during CI, but has to be manually run.

Issue #175, opened 2023-07-26
`CARGO_WORKSPACE_DIR` not set during runtime error when crates are exported


da-bao-jian:
When crates that have a dependency on `contracts` crate get [exported as packages](https://github.com/qi-protocol/4337-research/blob/6d27b246d5f464c85ffa489b6a5d00dfed8330bd/research/aa-bundler-runner/Cargo.toml#L10), during compilation, one would run into `CARGO_WORKSPACE_DIR` not set during runtime error. 

To work around this, I had to run `export CARGO_WORKSPACE_DIR=/my/specific/directory` before `cargo build` in the terminal. 

Are there ways to work around this in the script?

Vid201:
The way we handle smart contracts could be changed all together - we talked with @zsluedem about that once. At the moment, we are cloning the repo with solidity files - we could use `sol!` from `alloy-rs` instead (https://github.com/alloy-rs/core/tree/main/crates/sol-macro). This way the crate could become more "clean" as well.

PR #178, opened 2023-07-28
Change building env and thirdparty library for user to build as library

zsluedem:
Trying to fix this one https://github.com/Vid201/silius/issues/175

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,8 +23,8 @@ fetch-thirdparty:
 	git submodule update --init
 
 setup-thirdparty:
-	cd crates/contracts/account-abstraction && yarn install --frozen-lockfile --immutable && yarn compile && cd ../..
-	cd tests/bundler && yarn install --frozen-lockfile --immutable && yarn preprocess && cd ../..
+	cd crates/contracts/thirdparty/account-abstraction && yarn install --frozen-lockfile --immutable && yarn compile && cd ../../../..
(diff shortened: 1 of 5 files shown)


zsluedem:
@Vid201 Temporary fix now. Ultimately would use alloy.

Makefile
```diff
@@ -23,8 +23,8 @@ fetch-thirdparty:
 	git submodule update --init
 
 setup-thirdparty:
-	cd thirdparty/account-abstraction && yarn install --frozen-lockfile --immutable && yarn compile && cd ../..
-	cd thirdparty/bundler && yarn install --frozen-lockfile --immutable && yarn preprocess && cd ../..
+	cd crates/contracts/account-abstraction && yarn install --frozen-lockfile --immutable && yarn compile && cd ../..
```
Vid201, line 26: Shouldn't be `cd ../../..` at the end?

Vid201 (changes requested):
LGTM, this should solve the problem in the meantime

I am wondering if it's better to keep thirdparty folder -> meaning tests/thirdparty/bundler and crates/contracts/thirdparty/account-abstraction to indicate third party repos. Not sure if needed though

Makefile
zsluedem: yes. You are right

Vid201 (approved):
GJ

PR #183, opened 2023-08-06
Feat: send bundles to Flashbots

da-bao-jian:
RE: #165


zsluedem:
You could use `git rebase` to clean up your branch

bin/silius/src/silius.rs
```diff
@@ -56,11 +59,18 @@ pub struct Opt {
 
     #[clap(flatten)]
     pub bundler_opts: BundlerServiceOpts,
+
+    #[clap(long)]
+    pub build_fb_signer: Option<bool>,
+
+    #[clap(long, default_value = "eth-client")]
```
Vid201, line 66: We can do that will all the values defined for this param: https://github.com/Vid201/silius/blob/main/bin/silius/src/cli.rs#L51

eth-client, flashbots

crates/bundler/src/bundler.rs
```diff
@@ -1,40 +1,151 @@
...
 
-#[derive(Clone)]
+/// A type alias for the Ethereum Signer client
+type EthClientType<M> = Arc<SignerMiddleware<M, LocalWallet>>;
+
+/// The `SendBundleMode` determines whether to send the bundle to a Ethereum execution client or to Flashbots relay
+#[derive(Clone, Debug, PartialEq)]
+pub enum SendBundleMode {
+    /// Send the bundle to a Ethereum execution client
+    EthClient,
```
Vid201, line 29: Can we add to the enum Vec<String> to each variant>
EthClient(Vec<String>)
Flashbots(Vec<String>)

First contain list of eth execution client nodes, and the second flashbots relay endpoints. That way we can pass the string with enum and have less properties in struct Bundler

crates/bundler/src/bundler.rs
```diff
@@ -60,9 +188,58 @@ impl Bundler {
...
                 self.beneficiary,
             )
-            .tx
-            .clone();
-        tx.set_nonce(nonce).set_chain_id(self.chain.id());
+            .tx;
+
+        match self.chain.id() {
+            // Mumbai
+            80001u64 => {
```
Vid201, line 195: Same for chain id

crates/bundler/src/mock_relay.rs
```diff
@@ -0,0 +1,122 @@
...
+    SimulatedTransaction,
+};
+use jsonrpsee::{core::RpcResult, proc_macros::rpc};
+use std::sync::Arc;
+use std::time::Duration;
+
+pub const INIT_BLOCK: u64 = 17832062;
+
+#[rpc(server, namespace = "eth")]
+pub trait MockFlashbotsRelay {
```
Vid201, line 20: Is this needed only in tests?

Then we can add #[test]

Vid201 (changes requested):
This is looking good! left some smaller things to address

crates/bundler/src/bundler.rs
da-bao-jian: I see the point of using multiple relay endpoints, but I don't see why we need multiple execution client nodes. Each bundler should only use one execution client node. 

Is there a scenario where a bundler would use multiple execution client nodes?

crates/bundler/src/bundler.rs
```diff
@@ -1,40 +1,151 @@
...
-        Self {
-            wallet,
-            eth_client_address,
-            beneficiary,
-            entry_point,
-            chain,
+        send_bundle_mode: SendBundleMode,
+        relay_endpoints: Option<Vec<String>>,
+    ) -> anyhow::Result<Self> {
+        if !(chain.id() == 1 || chain.id() == 5 || chain.id() == 11155111)
```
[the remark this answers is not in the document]
da-bao-jian: By network id, do you mean chain id? Chain id is already included in the Bundler struct, and we can check which chain the bundler is on by `chain::id()`

Can you maybe give a pseudo code example or the context where this will be necessary? 

Thanks!

crates/bundler/src/mock_relay.rs
da-bao-jian: The mock_relay module has been marked as test
https://github.com/qi-protocol/silius/blob/24afb5ca8d1b095bf62fedc4b2b7eee14b2ce2f9/crates/bundler/src/lib.rs#L5

Will this suffice?

crates/bundler/src/bundler.rs
Vid201: Ok, you're right. Doesn't make a lot of sense. Let's leave it like that :+1:

crates/bundler/src/bundler.rs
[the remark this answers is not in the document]
Vid201: Option 1:

`chain.id() == ethers::types::Chain::Sepolia`

Option 2:

`chain::name() == "sepolia"`

I think it's better to have it like that because it's easier later to search in the code and find all specifics for some chain (for example gas things will be different for chain as well). For example if I search mainnet, I will find exactly things related to mainnet, but if I search "1" it will be a mess.

da-bao-jian:
Sorry took a little longer than expected. Mostly was resolving dependency conflicts and rebasing commits.

Vid201 (approved):

PR #185, opened 2023-08-06
Fix bundler rules

Vid201:
- set default values and remove some cli params in Makefile and README.md
- update storage access rule
- add additional bundling rule: Prevent bundling UserOps that access another UserOp's sender

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,10 +2,10 @@ build:
 	cargo build --release
 
 run-silius:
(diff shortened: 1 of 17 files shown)


zsluedem (approved):
LGTM. THis pr should fix the ci failing case, right?

zsluedem:
@Vid201 I merged it first since it is blocking others.

Issue #181, opened 2023-08-04
bundler being unresponsive after first userOp


stefanodecillis:
Hi,

I hope this problem is only on my side!
I deployed the docker image and it becomes unresponsive after handling one or two userOp. The log trace stops without any error, even if the application is not halting. 
Trying on different machines, the behavior is the same.

A typical UserOp I sent to the bundler is:

`{"method":"eth_sendUserOperation","params":[{"sender":"0x5481E5F531702E8fb0bCB7c98cBdb22e814F6Acd","nonce":"0x21","initCode":"0x","callData":"0x8d44ad620000000000000000000000002b87f2390e4aef1fc961027982804650fadfaf4000000000000000000000000000000000000000000000000000000000000000a0b08d656e32fc45f4c913e44936538785704ee65d06c3a86ef0b44238cca4b1db0000000000000000000000000000000000000000000000000000000064cccf6900000000000000000000000000000000000000000000000000000000000001e00000000000000000000000000000000000000000000000000000000000000104a5efb2350000000000000000000000005481e5f531702e8fb0bcb7c98cbdb22e814f6acd0000000000000000000000000000000000000000000000000000000000000040000000000000000000000000000000000000000000000000000000000000000100000000000000000000000000000000000000000000000000000000000000200000000000000000000000007a8a552e1305a631e2e1b44ba67b5d45a1a30497000000000000000000000000000000000000000000000000000000000000000100000000000000000000000000000000000000000000000000000000000000600000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000","callGasLimit":"0x0222c0","verificationGasLimit":"0x0249f0","preVerificationGas":"0xea60","maxFeePerGas":"0x3b9aca10","maxPriorityFeePerGas":"0x3b9aca00","paymasterAndData":"0x","signature":"0xbcaf5133b0ad85ada7295b966071b624df0f69782ac6e0729197ffdcd079220a010e04e85ef22b045c5e1b2f0eb863667af297e3c91dd5e91a7ab4710a9a67ca1b"},"0x5FF137D4b0FDCD49DcA30c7CF57E578a026d2789"],"id":46,"jsonrpc":"2.0"}`

The bundler handles correctly the userOp and sends the transaction on chain. After the first ones, it stops without any error log and the bundler becomes unresponsive.

Did you ever experience the same? 

The only changes I made to the docker file are:

```
[...]

PR #190, opened 2023-08-11
fix: entrypoint error deserialize bug

zsluedem:


Vid201 (approved):
LGTM! Do you think this can be a problem in the issue #181 ?

zsluedem:
> LGTM! Do you think this can be a problem in the issue #181 ?

No. That's another bug I found when I was trying to debug #181

PR #196, opened 2023-08-15
Update ci.yml

Vid201:
Bundler spec tests are being updated more often lately, so it's better to set this to specific commit.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -51,10 +51,12 @@ jobs:
       - name: Install Geth and solc
         run: |
           sudo add-apt-repository ppa:ethereum/ethereum
-          sudo add-apt-repository ppa:ethereum/ethereum-dev
           sudo apt-get update
           sudo apt-get install solc
-          sudo apt-get install ethereum
+          wget -c https://gethstore.blob.core.windows.net/builds/geth-linux-amd64-1.11.6-ea9e62ca.tar.gz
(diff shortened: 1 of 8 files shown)


zsluedem:
https://github.com/Vid201/silius/actions/runs/5872101953/job/15927412618?pr=196#step:13:94
The jsonrpc server didn;t start.

zsluedem:
The entrypoint address is changed based on the log. Dont know why.

.github/workflows/ci.yml
```diff
@@ -91,7 +92,7 @@ jobs:
 
       - run: pip install jq yq
 
-      - run: pdm install && pdm run update-deps
+      - run: pdm install && git submodule update --init --recursive && cd @account-abstraction && yarn && yarn compile &&  cd ../spec && yarn && yarn build
```
zsluedem, line 95: @Vid201 Finally locate the error is here. 

`pdm run update-deps` would interprete into `git submodule update --remote --init --recursive && cd @account-abstraction && yarn && yarn compile &&  cd ../spec && yarn && yarn build`

with `--remote` , submodule would update account-abstraction and bundler-spec repo to latest version.

Vid201: It's weird because nothing changed. I was thinking if the following can be a problem: each test setup geth instance and uses TempDir `test_geth` - after the test is finished this dir is deleted. Can this be a problem if tests are run async?

Issue #144, opened 2023-06-07
Handle min_balance and gas_factor bundler CMD args


Vid201:
Currently, these two parameters are ignored.

PR #199, opened 2023-08-19
chore: remove gas factor and update min balance

Vid201:
#144

### DIFF
--- a/crates/bundler/src/bundler.rs
+++ b/crates/bundler/src/bundler.rs
@@ -62,15 +62,16 @@ impl Bundler {
             .clone()
             .get_balance(self.wallet.signer.address(), None)
             .await?;
+        let beneficiary = if balance < self.min_balance {
+            self.wallet.signer.address()
+        } else {
+            self.beneficiary
+        };
 
         let mut tx: TypedTransaction = ep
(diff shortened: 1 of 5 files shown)


crates/bundler/src/bundler.rs
```diff
@@ -55,10 +58,19 @@ impl Bundler {
...
+
         let mut tx: TypedTransaction = ep
             .handle_ops(
                 uos.clone().into_iter().map(Into::into).collect(),
-                self.beneficiary,
+                if balance < self.min_balance {
+                    self.wallet.signer.address()
+                } else {
+                    self.beneficiary
+                },
```
zsluedem, line 73: Maybe putting some logs here so that people know when and who is being charged and funding.

zsluedem (changes requested):

crates/bundler/src/bundler.rs
Vid201: I am gonna add some logging, makes sense. Charged is being always the account derived from seed phrase that is passed in CLI, but beneficiary can be any address.

zsluedem (approved):
LGTM

PR #206, opened 2023-08-25
chore: remove online compiling

zsluedem:
I am having some network problems locally in my development env. I hope we could rely less on networking when compile stuff.


Vid201 (approved):
LGTM

PR #207, opened 2023-08-25
chore: ini ethers rev

shunkakinoki:
Pin the `ethers-solc` dep

### DIFF
--- a/crates/contracts/Cargo.toml
+++ b/crates/contracts/Cargo.toml
@@ -26,4 +26,4 @@ tokio = { workspace = true }
 [build-dependencies]
 anyhow = "1"
 ethers = { workspace = true, features = ["solc-full"] }
-ethers-solc = "2.0.8"
+ethers-solc = { workspace = true }
--- a/crates/contracts/Cargo.toml
+++ b/crates/contracts/Cargo.toml
@@ -26,4 +26,4 @@ tokio = { workspace = true }
 [build-dependencies]
 anyhow = "1"
 ethers = { workspace = true, features = ["solc-full"] }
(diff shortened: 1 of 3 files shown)


Vid201 (changes requested):
Hey, thanks for the PR. Make sense to change that, but I am not sure the best place for ethers-solc is in the workspace Cargo.toml since it's only used in the crate contracts. Let's change there the version and fix the conflicts - then it's good to merge!

shunkakinoki:
Updated! Thank you so much 🙌

crates/contracts/Cargo.toml
```diff
@@ -26,4 +26,4 @@ tokio = { workspace = true }
 [build-dependencies]
 anyhow = "1"
 ethers = { workspace = true, features = ["solc-full"] }
-ethers-solc = "2.0.8"
+ethers-solc = { git = "https://github.com/gakonst/ethers-rs", rev = "fa3017715a298728d9fb341933818a5d0d84c2dc" }
```
zsluedem, line 29: Could you please give more info on why 2.0.8 not working and what you are fixing? I don't understand stand the intention here

zsluedem (changes requested):

crates/contracts/Cargo.toml
shunkakinoki: Thank you for the review!

In my local machine, `2.0.9` introduces breaking changes and does not compile - also to keep in sync w/ the workspace version!

zsluedem: Would you mind post the compile error here?

Issue #191, opened 2023-08-11
Fix type conversion overflow problem


zsluedem:
1. type coversion

https://github.com/Vid201/silius/blob/33dc6e686fd2fff4b26e0c5ceadce20e3cf5e739/crates/uopool/src/utils.rs#L53-L86

There are a lot of `U256  -> u128 -> f64` type conversion which could possibly cause overflow problems. Use U256 to calculate all.

2. divide zero error 
https://github.com/Vid201/silius/blob/33dc6e686fd2fff4b26e0c5ceadce20e3cf5e739/crates/uopool/src/utils.rs#L85

`fee_per_gas` could be zero based on the user input. We should handle zero case

PR #214, opened 2023-09-25
Fix `calculate_pre_verification_gas` not to happen overflow problem

winor30:
### Description

The previous implementation had a potential for overflow. The logic has been modified to prevent this issue.

### Changes

- Removed the use of floating point calculations which might lead to inaccuracies and potential overflows.
- Replaced the previous calculation logic with a more straightforward and safer approach using `saturating_add`, `checked_div`, and other safe arithmetic operations.
- Added tests to ensure the new logic works correctly, especially in scenarios with large user operations and potential overflow situations.

### Relate
`1. type coversion` in https://github.com/Vid201/silius/issues/191

### DIFF
--- a/crates/uopool/src/utils.rs
+++ b/crates/uopool/src/utils.rs
@@ -62,25 +62,36 @@ impl Overhead {
     /// The pre-verification gas of the [UserOperation](UserOperation)
     pub fn calculate_pre_verification_gas(&self, uo: &UserOperation) -> U256 {
         let uo_pack = uo.pack();
-        let call_data: U256 = U256::from(
-            uo_pack
-                .deref()
-                .iter()
-                .map(|&x| {
-                    if x == 0 {
-                        self.zero_byte.as_u128()
-                    } else {
-                        self.non_zero_byte.as_u128()
-                    }
-                })
-                .sum::<u128>(),
-        );
-        let len_in_word = ((uo_pack.len() + 31) as f64) / 32_f64;
-        u256_from_f64_saturating(
-            (self.fixed.as_u128() as f64) / (self.bundle_size.as_u128() as f64)
-                + ((call_data + self.per_user_op).as_u128() as f64)
-                + (self.per_user_op_word.as_u128() as f64) * len_in_word,
-        )
+
+        let call_data = uo_pack.deref().iter().fold(U256::zero(), |acc, &x| {
+            let byte_cost = if x == 0 {
+                &self.zero_byte
+            } else {
+                &self.non_zero_byte
+            };
+            acc.saturating_add(*byte_cost)
+        });
+
+        let word_len = uo_pack.len() + 31;
+        let word_cost = self
+            .per_user_op_word
+            .saturating_mul(U256::from(word_len))
+            .checked_div(U256::from(32))
+            .unwrap_or_default()
+            .saturating_add(U256::from(if (word_len) % 32 > 0 { 1 } else { 0 }));
+
+        let fixed_divided_by_bundle_size = self
+            .fixed
+            .saturating_mul(U256::from(1_000_000))
+            .checked_div(self.bundle_size)
+            .unwrap_or_default()
+            .checked_div(U256::from(1_000_000))
+            .unwrap_or_default();
+
+        fixed_divided_by_bundle_size
+            .saturating_add(call_data)
+            .saturating_add(self.per_user_op)
+            .saturating_add(word_cost)
     }
 }
 
--- a/crates/uopool/src/utils.rs
+++ b/crates/uopool/src/utils.rs
@@ -317,6 +317,16 @@ pub mod tests {
         );
     }
 
+    #[test]
+    fn div_ceil_divisible_calculation() {
+        assert_eq!(div_ceil(U256::from(10), U256::from(2)), 5.into());
+    }
+
+    #[test]
+    fn div_ceil_no_divisible_calculation() {
+        assert_eq!(div_ceil(U256::from(10), U256::from(3)), 4.into());
+    }
+
     pub fn mempool_test_case<T>(mut mempool: T, not_found_error_message: &str)
     where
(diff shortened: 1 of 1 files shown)


crates/uopool/src/utils.rs
```diff
@@ -62,25 +62,36 @@ impl Overhead {
...
+        let word_cost = self
+            .per_user_op_word
+            .saturating_mul(U256::from(word_len))
+            .checked_div(U256::from(32))
+            .unwrap_or_default()
+            .saturating_add(U256::from(if (word_len) % 32 > 0 { 1 } else { 0 }));
+
+        let fixed_divided_by_bundle_size = self
+            .fixed
+            .saturating_mul(U256::from(1_000_000))
```
zsluedem, line 85: I am curious why 1_000_000 here.

crates/uopool/src/utils.rs
```diff
@@ -62,25 +62,36 @@ impl Overhead {
...
+            acc.saturating_add(*byte_cost)
+        });
+
+        let word_len = uo_pack.len() + 31;
+        let word_cost = self
+            .per_user_op_word
+            .saturating_mul(U256::from(word_len))
+            .checked_div(U256::from(32))
+            .unwrap_or_default()
+            .saturating_add(U256::from(if (word_len) % 32 > 0 { 1 } else { 0 }));
```
zsluedem, line 81: Why is this needed? You want to get the same result from old calculation, right?

crates/uopool/src/utils.rs
```diff
@@ -146,6 +157,91 @@ pub mod tests {
         assert_eq!(gas_oh.calculate_pre_verification_gas(&uo), 45340.into());
     }
 
+    #[test]
+    fn pre_verification_gas_calculation_with_large_user_operation() {
```
zsluedem, line 161: love it!

crates/uopool/src/utils.rs
```diff
@@ -146,6 +157,91 @@ pub mod tests {
...
+            sender: "0xAB7e2cbFcFb6A5F33A75aD745C3E5fB48d689B54"
+                .parse()
+                .unwrap(),
+            nonce: U256::max_value(),
+            init_code: Bytes::from(vec![255; 1024]), // Large init_code
+            call_data: Bytes::from(vec![255; 1024]), // Large call_data
+            call_gas_limit: U256::max_value(),
+            verification_gas_limit: U256::max_value(),
+            pre_verification_gas: U256::max_value(),
+            max_fee_per_gas: U256::max_value(),
```
zsluedem, line 203: It could be divided by zero if `max_fee_per_gas` is zero based on #191 . Would you mind also fixing it?

zsluedem (commented):
Very awesome work~! I really love your pr and want to merge it though I had some questions. 

https://github.com/Vid201/silius/blob/33dc6e686fd2fff4b26e0c5ceadce20e3cf5e739/crates/uopool/src/utils.rs#L78-L82
Do you think we overflow could also be possible here? Would you mind addressing it?

It is ok to refuse if you don't want to.

Thanks very much~!!

crates/uopool/src/utils.rs
```diff
@@ -94,9 +108,39 @@ impl Overhead {
...
+/// performs the division, and rounds up if there is a remainder.
+///
+/// # Examples
+///
+/// ```
+/// # use your_crate::U256; // Replace with the actual import
+/// let result = div_ceil(U256::from(10), U256::from(3));
+/// assert_eq!(result, U256::from(4));
+/// ```
+pub fn div_ceil(numerator: U256, denominator: U256) -> U256 {
```
winor30, line 132:
```suggestion
fn div_ceil(numerator: U256, denominator: U256) -> U256 {
```

crates/uopool/src/utils.rs
winor30: This was a scaling constant to prevent rounding of the division result.
However, 1_000_000 could round the quotient, so [`div_ceil`](https://github.com/Vid201/silius/pull/214/files#diff-4edf5911f99a49053c851d9441d21d3054641602def24869489b642956c5479bR132) was provided.
div_ceil is also used in other parts of the division.

crates/uopool/src/utils.rs
winor30: This process rounds up the decimal point when the result returns a decimal point.
If the result of `% 32` is non-zero, the decimal point is present and 1 is added.

But this has also been changed to use [div_ceil](https://github.com/Vid201/silius/pull/214/files#diff-4edf5911f99a49053c851d9441d21d3054641602def24869489b642956c5479bR122-R123).

crates/uopool/src/utils.rs
winor30: ok! i will fix it

winor30: Is it correct like this?
[`6e3b6cd` (#214)](https://github.com/Vid201/silius/pull/214/commits/6e3b6cde063f5a72c3d3ed6cbbc51f74f35eb3f8)

zsluedem (approved):
Very good work~! Thanks for addressing all the comments and the additional issues~! 

@Vid201 do you think this is good?

Vid201:
> Very good work~! Thanks for addressing all the comments and the additional issues~!
> 
> @Vid201 do you think this is good?

Looks good to me! GJ @winor30 

Can we just squash all commits when merging @zsluedem

zsluedem:
@winor30 Please squash the commits into one. Then I will merge it. Thanks very much~!

winor30:
@Vid201 @zsluedem thank you for your review!

@zsluedem 
> Please squash the commits into one. Then I will merge it. Thanks very much~!

I could squash the commits myself, but I thought it would be better to have a squash merge.
like below
image

zsluedem:
@winor30 Thanks. I didn't know github got this feature actually. Anyway, really appreciate your help.

PR #216, opened 2023-09-26
chore: support both http and ws when connecting to eth execution client

Vid201:


zsluedem (approved):
Great. Looks easier than I expected.

PR #220, opened 2023-10-06
Listen for new blocks and remove user operations

Vid201:
Bundler now subscribes to new blocks and removes all user operations included on-chain.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,12 +21,14 @@ repository = "https://github.com/Vid201/silius"
 rust-version = "1.71.1"
 
 [workspace.dependencies]
+async-stream = "0.3.5"
 async-trait = "0.1"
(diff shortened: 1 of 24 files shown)


Cargo.toml
```diff
@@ -14,19 +14,21 @@ default-members = ["bin/silius"]
 
 [workspace.package]
 authors = ["Vid Kersic <vid.kersic@yahoo.com>"]
-version = "0.2.0-alpha"
+version = "0.3.0-alpha"
```
zsluedem, line 17: Next time, We should bump the version first and release.

zsluedem (approved):
LGTM

Cargo.toml
Vid201: Yeah, that was my bad, I forgot that. Will do it next time

zsluedem: Not a big deal right now. Not blaming here. I made the same mistake sometimes. Additionally, we could use bot to release.

PR #221, opened 2023-10-07
Improve json rpc error message

zsluedem:
```
res: "{\"jsonrpc\":\"2.0\",\"error\":{\"code\":-32603,\"message\":\"Unknown error when simulating validation on entry point. Error message: JsonRpcError(JsonRpcError { code: -32015, message: \\\"VM execution error.\\\", data: Some(String(\\\"Reverted 0x220266b600000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000040000000000000000000000000000000000000000000000000000000000000001e41413430206f76657220766572696669636174696f6e4761734c696d69740000\\\")) })\"},\"id\":1}"
```

We could see error message like above in estimateGas. The reverted message could be deserialize to Failop with more concrete message. We should provide that to user with better experience.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4340,13 +4340,13 @@ dependencies = [
 
 [[package]]
 name = "regex"
-version = "1.9.5"
+version = "1.9.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


Vid201 (approved):
LGTM!

PR #224, opened 2023-10-11
Replace dashmap with hashmap

zsluedem:
Replace dashmap with hashmap to see whether dashmap is the problem causing data race in #181


Vid201:
What's the status on this one @zsluedem ?

zsluedem:
> What's the status on this one @zsluedem ?

Hm.. Need more tests on this. Close it for now.

PR #247, opened 2023-11-26
Refactor to abstract storage engine and fix bugs

zsluedem:
This is a very big pr for several issues
1. https://github.com/silius-rs/silius/issues/243
2. fix https://github.com/silius-rs/silius/issues/230 and https://github.com/silius-rs/silius/issues/246
3. Solve https://github.com/silius-rs/silius/issues/246 by separate locks for different component
4. Provide args to use memory or database mempool

### DIFF
--- a/bin/silius/src/bundler.rs
+++ b/bin/silius/src/bundler.rs
@@ -4,15 +4,25 @@ use crate::{
 };
 use alloy_chains::Chain;
 use ethers::{providers::Middleware, types::Address};
+use parking_lot::RwLock;
+use silius_contracts::EntryPoint;
 use silius_grpc::{
     bundler_client::BundlerClient, bundler_service_run, uo_pool_client::UoPoolClient,
     uopool_service_run,
 };
 use silius_primitives::{
     bundler::SendBundleMode,
-    consts::{flashbots_relay_endpoints, p2p::DISCOVERY_SECRET_FILE_NAME},
+    consts::{entry_point, flashbots_relay_endpoints, p2p::DISCOVERY_SECRET_FILE_NAME},
+    consts::{
+        p2p::DB_FOLDER_NAME,
+        reputation::{
+            BAN_SLACK, MIN_INCLUSION_RATE_DENOMINATOR, MIN_UNSTAKE_DELAY, THROTTLING_SLACK,
+        },
+    },
     provider::BlockStream,
-    Wallet,
+    reputation::ReputationEntry,
+    simulation::CodeHash,
+    UserOperation, UserOperationHash, Wallet,
 };
 use silius_rpc::{
     debug_api::{DebugApiServer, DebugApiServerImpl},
@@ -20,7 +30,19 @@ use silius_rpc::{
     web3_api::{Web3ApiServer, Web3ApiServerImpl},
     JsonRpcServer, JsonRpcServerType,
 };
-use std::{collections::HashSet, future::pending, net::SocketAddr, str::FromStr, sync::Arc};
+use silius_uopool::{
+    init_env,
+    validate::validator::{new_canonical, new_canonical_unsafe},
+    CodeHashes, DatabaseTable, EntitiesReputation, Mempool, Reputation, UserOperations,
+    UserOperationsByEntity, UserOperationsBySender, WriteMap,
+};
+use std::{
+    collections::{HashMap, HashSet},
+    future::pending,
+    net::SocketAddr,
+    str::FromStr,
+    sync::Arc,
+};
 use tracing::info;
 
 pub async fn launch_bundler<M>(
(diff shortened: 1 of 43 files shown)


crates/uopool/src/database/mempool.rs
```diff
@@ -1,383 +1,252 @@
...
-    /// * `Ok(Option<UserOperation>)` - The user operation if it exists.
-    /// * `Err(DBError)` - The database error.
-    fn get(&self, uo_hash: &UserOperationHash) -> Result<Option<UserOperation>, DBError> {
+    fn remove_by_uo_hash(&mut self, uo_hash: &UserOperationHash) -> Result<bool, MempoolError> {
         let uo_hash_wrap: WrapUserOperationHash = (*uo_hash).into();
 
-        let tx = self.env.tx()?;
-        let res = tx.get::<UserOperations>(uo_hash_wrap)?;
+        let tx = self.env.tx_mut()?;
+        let ori = tx.get::<UserOperations>(uo_hash_wrap.clone())?;
```
Vid201, line 41: ori?

crates/uopool/src/validate/mod.rs
```diff
@@ -51,47 +50,107 @@ pub enum UserOperationValidatorMode {
...
+        X: UserOperationAddrAct,
+        Z: UserOperationCodeHashAct,
+        H: HashSetOp,
+        R: ReputationEntryOp,
+    {
+        Ok(())
+    }
 }
 
+sanity_check_impls! { A }
```
Vid201, line 143: Some clarification comments

crates/uopool/src/validate/mod.rs
```diff
@@ -51,47 +50,107 @@ pub enum UserOperationValidatorMode {
...
 /// The [UserOperation](UserOperation) sanity check helper trait.
-pub struct SanityHelper<'a, M: Middleware + 'static, P, R, E>
-where
-    P: Mempool<UserOperations = VecUo, CodeHashes = VecCh, Error = E> + Send + Sync,
-    R: Reputation<ReputationEntries = Vec<ReputationEntry>, Error = E> + Send + Sync,
-{
-    mempool: &'a MempoolBox<VecUo, VecCh, P, E>,
-    reputation: &'a ReputationBox<Vec<ReputationEntry>, R, E>,
-    entry_point: EntryPoint<M>,
+pub struct SanityHelper<'a, M: Middleware + 'static> {
```
Vid201, line 71: I think we don't need SanityHelper anymore since entry_point and chain can be acccessed through mempool

crates/uopool/src/mempool.rs
```diff
@@ -1,164 +1,353 @@
...
+}
+impl<T> UserOperationAddrAct for T where
+    T: AddRemoveUserOpHash + UserOperationAddrOp + ClearOp + Send + Sync
+{
+}
+
+pub trait UserOperationCodeHashAct: UserOperationCodeHashOp + ClearOp + Send + Sync {}
+impl<T> UserOperationCodeHashAct for T where T: UserOperationCodeHashOp + ClearOp + Send + Sync {}
+
+pub struct Mempool<T, Y, X, Z>
```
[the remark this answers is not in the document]
Vid201: I'm just curious - does this pattern have a name? Just interested if this is some standardised approach

Vid201 (commented):
Generally looks good!

Left some comments

crates/uopool/src/validate/mod.rs
zsluedem: Mempool doesn't provide this now after the refactoring.

examples/storage/examples/database.rs
```diff
@@ -1,48 +1,24 @@
...
-use std::{env, str::FromStr, sync::Arc};
+use silius_uopool::{init_env, DatabaseTable, UserOperationOp, UserOperations, WriteMap};
+use std::{env, sync::Arc};
 use tempdir::TempDir;
 
 #[tokio::main]
 async fn main() -> eyre::Result<()> {
     //  uopool needs connection to the execution client
-    if let Ok(provider_url) = env::var("PROVIDER_URL") {
+    if let Ok(_) = env::var("PROVIDER_URL") {
```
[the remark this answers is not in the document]
zsluedem: I would do as you ask although I don't think it is neccessary

bin/silius/src/bundler.rs
```diff
@@ -161,29 +183,226 @@ where
...
-    .await?;
-
-    info!(
-        "Started uopool gRPC service at {:?}:{:?}",
-        args.uopool_addr, args.uopool_port
+    let entrypoint_api = EntryPoint::new(
+        eth_client.clone(),
+        Address::from_str(entry_point::ADDRESS).expect("address should be valid"),
     );
+    match (args.uopool_mode, args.use_memory) {
```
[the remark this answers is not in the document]
zsluedem: This is the disadvantage of the refactoring. When things comes to be totally type concrete, the compiler would complain a lot about it which leads to the highly duplicate codes here. It is not easy to do that unless we introduce dynamic dispatch for Mempool , Reputation and validator. 

I suggest we leave it here right now and I know it is ugly. We could optimize the codes here when I finished dynamic dispatch bench mark and it doesn't cost too much performance penalty. If the performance is an issue for dynamic dispatch, we could only introduce some macro to make codes more readible here.

crates/uopool/src/mempool.rs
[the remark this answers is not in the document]
zsluedem: I am actually sure what name should be used actually. This is like a common practice for trait system (mostly for function programming language not oop). The goal is to create minimize method and compose these methods to make a program. 
https://scg.unibe.ch/research/traits maybe the lists here are helpful.

bin/silius/src/cli/args.rs
```diff
@@ -22,6 +22,12 @@ use std::{
     path::PathBuf,
 };
 
+#[derive(ValueEnum, Debug, Clone)]
+pub enum StorageType {
+    MDBX,
```
Vid201, line 27: Let's rename this to database - too much details for someone using Silius

bin/silius/src/cli/args.rs
```diff
@@ -75,6 +81,11 @@ pub struct UoPoolArgs {
     #[clap(long)]
     pub datadir: Option<ExpandedPathBuf>,
 
+    /// If true, the uopool will use an in-memory database.
```
Vid201, line 84: Update comment

Vid201 (approved):
Can be merged after resolve last two new comments

PR #267, opened 2023-12-26
feat: support all available chains

zsluedem:


Vid201 (commented):
I see a problem with automatically adding support for every chain, specially tied to L2s: the gas for bundle is calculated differently on L2s (for example on Optimism and Arbitrum, and even between those two L2s there are differences). Users running Silius for the first time on some EVM chain could end up in weird behaviour.

That's way I think it's better to manually add supported chains on which the Silius was tested successfully.

zsluedem:
> I see a problem with automatically adding support for every chain, specially tied to L2s: the gas for bundle is calculated differently on L2s (for example on Optimism and Arbitrum, and even between those two L2s there are differences). Users running Silius for the first time on some EVM chain could end up in weird behaviour.
> 
> That's way I think it's better to manually add supported chains on which the Silius was tested successfully.

Maybe we could just put a big warning for user that some of the chains are not officially supported and use at your own risk?

crates/primitives/src/constants.rs
```diff
@@ -80,7 +80,6 @@ pub mod flashbots_relay_endpoints {
 /// Supported chains
 pub mod supported_chains {
     pub const SUPPORTED_CHAINS: [u64; 5] = [MAINNET, GOERLI, SEPOLIA, DEV, MUMBAI];
```
Vid201, line 82: We don't need this const anymore, right?

Vid201:
> > I see a problem with automatically adding support for every chain, specially tied to L2s: the gas for bundle is calculated differently on L2s (for example on Optimism and Arbitrum, and even between those two L2s there are differences). Users running Silius for the first time on some EVM chain could end up in weird behaviour.
> > That's way I think it's better to manually add supported chains on which the Silius was tested successfully.
> 
> Maybe we could just put a big warning for user that some of the chains are not officially supported and use at your own risk? @Vid201

Ok, let's do that. Then let's add this in the README, in the section supported chains. We can say it was tested on the chains in the table, but you can try on any other chain if you want.

Vid201 (approved):

PR #270, opened 2023-12-30
chore: update docker build, ci

Vid201:


Dockerfile
```diff
@@ -1,40 +1,55 @@
...
 RUN apk add openssl1.1-compat
 
-COPY --from=builder /silius/target/release/silius /usr/local/bin/silius
+# Copy silus binary
+COPY --from=builder /app/silius /usr/local/bin/silius
 
+# Copy licenses
+COPY LICENSE-* ./
+
+# Expose ports
```
zsluedem, line 52: We could build "locally" and just copy the target into the image. In this case, docker image is much smaller.

Dockerfile
```diff
@@ -1,40 +1,55 @@
-# build
-FROM ubuntu:18.04 AS builder
+FROM lukemathwalker/cargo-chef:latest-rust-1 AS chef
```
zsluedem, line 1: We could build "locally" and just copy the target into the image. In this case, docker image is much smaller.

Vid201: That's true. But I would keep the building inside Docker so someone who can't build Silius locally (due to unsupported architecture) can build it in Docker.

zsluedem (approved):
LGTM

PR #274, opened 2024-01-03
chore: get_logs with block depth

zsluedem:


crates/mempool/src/uopool.rs
```diff
@@ -509,10 +511,12 @@ where
         uo_hash: &UserOperationHash,
     ) -> eyre::Result<Option<(UserOperationEventFilter, LogMeta)>> {
         let mut event: Option<(UserOperationEventFilter, LogMeta)> = None;
+        let latest_block = self.entry_point.eth_client().get_block_number().await?;
         let filter = self
             .entry_point
             .entry_point_api()
             .event::<UserOperationEventFilter>()
+            .from_block(latest_block - FILTER_MAX_DEPTH)
```
zsluedem, line 519: For some very short block chains, it is not possible to catch the receipt with shallow-depth search. Increase the depth here to make get the receipt possible. But it is still not an ideal solution

Vid201 (approved):

PR #275, opened 2024-01-09
doc: add public endpoint

zsluedem:


Vid201 (changes requested):
Looking good :fire: 

I think we can merge this list and supported networks table:  add another column public endpoint.

README.md
```diff
@@ -74,12 +74,31 @@ docker run --net=host -v ./bundler-spec-tests/keys/0xf39Fd6e51aad88F6F4ce6aB8827
...
+| Ethereum Holesky | https://rpc.silius.xyz/api/v1/chain/ethereum-holesky |
+| Polygon | :soon: |
+| Polygon Mumbai | https://rpc.silius.xyz/api/v1/chain/polygon-mumbai |
+| Linea | :soon: |
+| Linea Goerli | https://rpc.silius.xyz/api/v1/chain/linea-goerli |
+| Optimism | :soon: |
+| Optimism Goerli | https://rpc.silius.xyz/api/v1/chain/optimism-goerli |
+| Optimism Sepolia | https://rpc.silius.xyz/api/v1/chain/optimism-sepolia |
+| Arbitrum | :soon: |
+| Arbitrum Goerli | https://rpc.silius.xyz/api/v1/chain/optimism-goerli |
```
Vid201, line 93: Wrong URL?

Vid201 (changes requested):
Let's have two columns - what is supported and public endpoint url? Maybe we support some network, but don't have public endpoint right away

zsluedem:
> Let's have two columns - what is supported and public endpoint url? Maybe we support some network, but don't have public endpoint right away

It is kind of wired that some networks are not supported but have public endpoints.....

Vid201:
> > Let's have two columns - what is supported and public endpoint url? Maybe we support some network, but don't have public endpoint right away
> 
> It is kind of wired that some networks are not supported but have public endpoints.....

Hmm, it was like that till now. We can always leave it like that if you feel it's better.

Vid201 (approved):

PR #277, opened 2024-01-13
chore: update supported chains

Vid201:


crates/bundler/src/bundler.rs
```diff
@@ -204,9 +203,9 @@ where
         let mut tx: TypedTransaction =
             ep.handle_ops(uos.clone().into_iter().map(Into::into).collect(), beneficiary).tx;
 
-        match self.chain.id() {
+        match Chain::from_id(self.chain.id()).named() {
             // Mumbai
-            supported_chains::MUMBAI => {
+            Some(NamedChain::PolygonMumbai) => {
```
zsluedem, line 208: Maybe this is not related to the pr but I was wondering why polygon is different when I was setting up the public endpoints. Any ideas?

Vid201: Hmm, not sure about that, we should investigate that

PR #280, opened 2024-01-18
fix: uo by sender not removed bug

zsluedem:
This is a bug which caused people would see not enough stake error when they sent more than 4 userops.


Vid201 (approved):
Nice catch!

PR #289, opened 2024-02-23
feat: new estimate method

zsluedem:


crates/mempool/src/estimate.rs
```diff
@@ -0,0 +1,272 @@
...
+    UserOperationRevertReasonFilter,
+};
+use silius_primitives::UserOperationSigned;
+use std::str::FromStr;
+
+const FALL_BACK_BINARY_SEARCH_CUT_OFF: u64 = 30000;
+const BASE_VGL_BUFFER: u64 = 25;
+const MAX_CALL_GAS_LIMIT: u64 = 18_000_000;
+const MAX_RETRY: u64 = 7;
+const NON_ZERO_GAS: u64 = 12100; // should be different based on diferrent chain
```
Vid201, line 22: Is this value chain specific? Then we should move this value to the chain struct and set different values for different chains

Vid201: We probably need chain spec thing to handle different values on different chains

Vid201 (changes requested):
LGTM!

We need to handle some L2s (starting with Arbitrum and Optimism) as well, example: https://github.com/etherspot/skandha/tree/master/packages/params/src/gas-estimation

crates/mempool/src/estimate.rs
zsluedem: I am not sure what other chains the value is. It may be different or not.It is good to have it what it is right now.

crates/contracts/src/error.rs
```diff
@@ -129,23 +133,25 @@ impl EntryPointError {
...
+// case revert with `require` error would ends up with error event signature `0x08c379a0`
+// we need to handle it manually
+pub fn decode_revert_string(data: Bytes) -> Option<String> {
+    let (error_sig, reason) = data.split_at(4);
+    if error_sig == [0x08, 0xc3, 0x79, 0xa0] {
+        <String as AbiDecode>::decode(reason).ok()
+    } else {
+        None
+    }
+}
```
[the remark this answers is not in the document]
zsluedem: It should go into rustfmt.toml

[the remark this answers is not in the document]
zsluedem: Anyway, it is none of the goal of this pr

crates/mempool/src/estimate.rs
Vid201: Ok, got it. But we should still handle the Optimism and Arbitrum differently - L1 gas and L2 gas separately (as other bundlers are handling)

zsluedem: Sorry. We could not do it easily right now. It is another big task. Would you mind merging this first ? Created an issue for that https://github.com/silius-rs/silius/issues/290.

Vid201 (approved):

PR #299, opened 2024-04-05
chore: p2p peer management and discovery mechanism

Vid201:
Peer management is now better handled and discovery problems are fixed.

------------------------------

ty lighthouse


zsluedem (approved):
LGTM

PR #314, opened 2024-06-14
feat: support loading private key

malm0d:
To load using private key, we can now use the `--private-key` option. And if we want to load with private key using flashbots send bundle mode, we can pass another option `--flashbots-private-key`. An addition option is required for using flashbots mode while loading with private key, since the private key only maps to a single address without hierarchical derivation.

Sample command to load using private key:
`cargo run --release -- node --eth-client-address http://127.0.0.1:8545 --private-key 4c5e5d30… --beneficiary 0xf39Fd6e5… --entry-points 0x5FF137D4… --http --ws`

Sample command to load using private key with flashbots mode:
`cargo run --release -- node --eth-client-address http://127.0.0.1:8545 --private-key 4c5e5d30… --flashbots-private-key df218be...  --send-bundle-mode flashbots --beneficiary 0xf39Fd6e5… --entry-points 0x5FF137D4… --http --ws`

- The `--mnemonic-file` and `--private-key` options are mutually exclusive, one and only one can be used. 
- Similarly, `--flashbots-private-key` cannot be used in the same command as `--mnemonic-file`, and it cannot be used alone without the `--private-key` option; but it can be optional when `--private-key` is used.
- If `--send-bundle-mode` is not `flashbots` and `--flashbots-private-key` is provided, `--flashbots-private-key` will be ignored.

Additional considerations:
I added a `value-parser` for the `--private-key` and `--flashbot-private-key` arguments, and the current implementation only checks if there are 64 characters and if each char is an ASCII hexadecimal digit. Am wondering if this is sufficient or if theres another better way to do this:
```
pub fn validate_private_key(hex_string: &str) -> Result<String, String> {
    let chars = hex_string.chars();

    if chars.clone().count() != 64 {
        return Err(format!("{hex_string} is not a valid private key"));
    }

    for c in chars {
        if !c.is_ascii_hexdigit() {
            return Err(format!("{hex_string} is not a valid hexadecimal string"));
        }
    }

    Ok(String::from(hex_string))
}
```

### DIFF
--- a/bin/silius/src/bundler.rs
+++ b/bin/silius/src/bundler.rs
@@ -121,15 +121,41 @@ where
     let chain_conn = Chain::from(chain_id);
 
     let wallet: Wallet;
-    if args.send_bundle_mode == SendStrategy::Flashbots {
-        wallet = Wallet::from_file(args.mnemonic_file.into(), chain_id, true)
-            .map_err(|error| eyre::format_err!("Could not load mnemonic file: {}", error))?;
-        info!("Wallet Signer {:?}", wallet.signer);
-        info!("Flashbots Signer {:?}", wallet.flashbots_signer);
+
+    if let Some(mnemonic_file) = args.mnemonic_file {
+        if args.send_bundle_mode == SendStrategy::Flashbots {
+            wallet = Wallet::from_file(mnemonic_file.into(), chain_id, true)
+                .map_err(|error| eyre::format_err!("Could not load mnemonic file: {}", error))?;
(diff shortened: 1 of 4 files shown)


Vid201 (changes requested):
This PR looks really good and clean, thanks!

I would just remove the check `validate_private_key` because the program would terminate either way if there is a wrong format/key - when converting to wallet in `wallet.rs`.

Vid201 (approved):
Merging this now, thanks for contributing!!

PR #320, opened 2025-02-06
feat: add direct usage example without gRPC

### LINKED ISSUE #217: Example of Silius components without gRPC
At the moment, binary creates three gRPC servers (bundling, uopool, RPC) that communicates over gRPC (look crate gRPC). It would be cool to have example that doesn't use gRPC (use crates bundler, uopool, rpc directly).

TobieTom:
Add example demonstrating direct usage of Silius components without gRPC communication.

Changes:
- Add examples/direct/main.rs with single binary implementation
- Add examples/direct/Cargo.toml with dependencies
- Add examples/Cargo.toml for example configuration

This example shows how to use bundler, mempool and RPC components directly in memory, providing a simpler deployment option when service isolation isn't needed.

Resolves #217

### DIFF
--- a/examples/direct/Cargo.toml
+++ b/examples/direct/Cargo.toml
@@ -1,0 +1,25 @@
+[package]
+name = "silius-direct-example" 
+version = "0.5.0-alpha"
+edition = "2021"
+
+[[bin]]
+name = "direct"
+path = "main.rs"
+
+[dependencies]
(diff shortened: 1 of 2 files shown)


Vid201:
Hey @TobieTom , thanks for opening a PR. I will take a look at it as soon as I can !

